Issue #5, opened 2021-11-05
Do we need addons to be sets of files or can simplify to be one file?


tigrannajaryan:
Addons are defined as a set of downloadable files. Can we simplify and make each addon a single downloadable file (which internally can be a tarball if needed)? Similarly the agent package, can we simplify it to be one downloadable file?

tigrannajaryan:
Opened a draft PR to show what it would look like: https://github.com/open-telemetry/opamp-spec/pull/29

I think it is worth doing this simplification.

PR #29, opened 2021-11-16
Change addons and agent packages to only contain one downloadable file

tigrannajaryan:
Resolves https://github.com/open-telemetry/opamp-spec/issues/5

This simplifies the protocol a bit and does not prevent the Agents to
have addons or packages that are actually composed of multiple files (which
can be stored in e.g. a zip or tar downdloadable file).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1533,9 +1533,9 @@ message AgentRemoteConfig {
 <h2 id="addons">Addons</h2>
 
 
-An Addon is a collection of named files. The content of the files, functionality
-provided by the addons, how they are stored and used by the Agent side is agent
-type-specific and is outside the concerns of the OpAMP protocol.
+An Addon has a name and content stored in a file. The content of the file,
+functionality provided by the addons, how they are stored and used by the Agent
+side is agent type-specific and is outside the concerns of the OpAMP protocol.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1787,86 +1777,57 @@ message has the following structure:
...
 
-
 The message has the following structure:
 
-
 ```protobuf
 message DownloadableFile {
-    string download_url = 1;
-    bytes hash = 2;
+    string file_name = 1;
```
tigrannajaryan, line 1803: Do we need the file name? What if we remove it? For addons we have the addon name to use the name of the local file to download to. For agent packages the agent is free to use a hard-coded name. It is not clear what we gain by having the file name in the message.

pmm-sumo: I agree that `file_name` seems to be providing little value. If we need the original file name for whatever purpose then we should be also able to extract it from the  `download_url` or [Content-Disposition](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition) header

tigrannajaryan: Removed.

pmm-sumo (approved):

djaglowski (approved):

Issue #25, opened 2021-11-15
Define how future extensions for the protocol capabilities will work


tigrannajaryan:
We need to define how the protocol can evolve and how new capabilities can be defined. Ideally we would want this evolution to be non-breaking, so that interoperability between new and old versions is preserved.

This is also related to the [mechanism which should be used](https://github.com/open-telemetry/opamp-spec/issues/7) when a particular currently defined capability is not supported by the agent or the server. Perhaps we can use the same mechanism for detection of both current and future capabilties.

Issue #7, opened 2021-11-09
Clarify how the protocols should work when a particular capability is not supported


tigrannajaryan:
The spec does not tell how the protocol works when the client or the serve don't support a particular capability (e.g. addons are not supported).

Need to clarify it.

This is also related to how we plan to support addition of new features to the protocol in the future. It may be that we can use the same capability signalling mechanism to express that fact that a particular feature is not supported by the peer, including features that are currently already in the spec or the features that we may add to the spec in the future.

PR #33, opened 2021-11-17
Add ability to indicate Capabilities in the protocol

tigrannajaryan:
Resolves https://github.com/open-telemetry/opamp-spec/issues/7
Resolves https://github.com/open-telemetry/opamp-spec/issues/25

This change clarifies how the protocols should work when a particular
capability is not supported. It also defines how future extensions
for the protocol capabilities will work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -237,7 +237,7 @@ message DataForAgent {
     AddonsAvailable addons_available = 3;
     AgentPackageAvailable agent_package_available = 4;
     Flags flags = 5;
-    Capabilities capabilities = 6;
+    ServerCapabilities capabilities = 6;
 }
 ```
 
@@ -307,7 +307,7 @@ be set in the first DataForAgent sent by the Server and MAY be omitted in
 subsequent DataForAgent messages by setting it to Unspecified value.
 
 ```protobuf
-enum Capabilities {
+enum ServerCapabilities {
   // The capabilities field is unspecified.
   Unspecified = 0;
   // The Server can accept status reports. This bit MUST be set, since all Server
@@ -513,7 +513,7 @@ message StatusReport {
     EffectiveConfig effective_config = 2;
     RemoteConfigStatus remote_config_status = 3;
     bytes server_provided_all_addons_hash = 4;
-    Capabilities capabilities = 5;
+    AgentCapabilities capabilities = 5;
 }
 ```
 
@@ -568,7 +568,7 @@ be set in the first StatusReport sent by the Agent and MAY be omitted in
 subsequent StatusReport messages by setting it to Unspecified value.
 
 ```protobuf
-enum Capabilities {
+enum AgentCapabilities {
     // The capabilities field is unspecified.
     Unspecified = 0;
     // The Agent can report status. This bit MUST be set, since all Agents MUST
--- a/README.md
+++ b/README.md
@@ -237,6 +237,7 @@ message DataForAgent {
     AddonsAvailable addons_available = 3;
     AgentPackageAvailable agent_package_available = 4;
     Flags flags = 5;
+    Capabilities capabilities = 6;
 }
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -283,15 +284,55 @@ enum Flags {
...
+
+Bitmask of flags defined by Capabilities enum. All bits that are not defined in
+Capabilities enum MUST be set to 0 by the Server. This allows extending the
+protocol and the Capabilities enum in the future such that old Servers
+automatically report that they don't support the new capability. This field MUST
+be set in the first DataForAgent sent by the Server and MAY be omitted in
+subsequent DataForAgent messages by setting it to Unspecified value.
+
+```protobuf
+enum Capabilities {
```
djaglowski, line 310: Should we differentiate the two enums?
```suggestion
enum ServerCapabilities {
```

README.md
```diff
@@ -516,6 +558,54 @@ The server SHOULD compare this hash to the aggregate hash of all addons that it
...
+
+Bitmask of flags defined by Capabilities enum. All bits that are not defined in
+Capabilities enum MUST be set to 0 by the Agent. This allows extending the
+protocol and the Capabilities enum in the future such that old Agents
+automatically report that they don't support the new capability. This field MUST
+be set in the first StatusReport sent by the Agent and MAY be omitted in
+subsequent StatusReport messages by setting it to Unspecified value.
+
+```protobuf
+enum Capabilities {
```
djaglowski, line 571: Corresponding to above comment
```suggestion
enum AgentCapabilities {
```

README.md
tigrannajaryan: They are differentiated by being nested in the corresponding message. But I like your suggestion, it makes things more clear.

README.md
tigrannajaryan: Done

djaglowski (approved):

tigrannajaryan:
@pmm-sumo I think you raised this question. Please see if this PR answers it.

README.md
```diff
@@ -283,15 +284,55 @@ enum Flags {
...
+  // The Server can offer Addons.
+  OffersAddons                   = 0x00000008;
+  // The Server can accept Addon status.
+  AcceptsAddonsStatus            = 0x00000010;
+  // The Server can offer packages to install.
+  OffersAgentPackage             = 0x00000020;
+  // The Server can accept the installation status of the package.
+  AcceptsAgentPackageStatus      = 0x00000040;
+  // The Server can offer connection settings.
+  OffersConnectionSettings       = 0x00000080;
```
pmm-sumo, line 329: AgentCapabilities have `ReportOwnTraces` at this position and then a bunch of other capabilities. Should we keep `ServerCapabilities` in sync with `AgentCapabilities`?

pmm-sumo: Now that I think about and look at the relationship from both ends, I see following:

* when server _offers_ something to the agent, it just advertises that certain information will be sent in the message; I don't think it changes anything (any of the "Offers" messages); 
* when server _accepts_ something from the agent, agent could pull that information and include it in the message
* when agent _reports_ something to the server, server can anticipate that certain data will be sent, which means it might e.g. wait for statuses before sending a message to the agent
* when agent _accepts_ something from the server, server can use this information to construct the message (e.g. if no addons are accepted, it might skip sending addons)

I might be missing something, but I am not sure how any of the server Offers* information can be practically used by the agent?

tigrannajaryan: > I might be missing something, but I am not sure how any of the server Offers* information can be practically used by the agent?

Exactly. You described it better than I could.

The only use that I can think of is logging on the Agent side so that when troubleshooting it is clear _why_ we are not seeing the offer from the Server. I don't know if we care about knowing this at a finer detail than just "Server can offer connection settings".

pmm-sumo: Got it. I think it won't hurt much to keep it in the specs.

Another thing I am wondering about though is if server should provide following: `AcceptsAgentTraces`, `AcceptsAgentMetrics`, `AcceptsAgentLogs`. I believe that while own telemetry setup would be provided by server, having such information might be still useful for the agent in certain scenarios (e.g. perhaps agent would fallback to alternative config or so)

tigrannajaryan: > Another thing I am wondering about though is if server should provide following: `AcceptsAgentTraces`, `AcceptsAgentMetrics`, `AcceptsAgentLogs`.

It would be kind of misleading though, since it is not the OpAMP server that accepts traces/metrics/logs, it is probably a different server (telemetry backend), the credentials of which are offered via ConnectionSettings message. I think the ConnectionSettings message already conveys that information.

>I believe that while own telemetry setup would be provided by server, having such information might be still useful for the agent in certain scenarios (e.g. perhaps agent would fallback to alternative config or so when server does not accept some of the own telemetry signals)

The Agent can already make this decision based on the _absence_ of the particular class of ConnectionSettings offer from the Server (e.g. if ConnectionSettingsOffers.own_traces field is unset by the Server). I think the additional capabilities bit is not necessary for that.

pmm-sumo (approved):

tigrannajaryan:
Let's merge this. We may need to tweak the capabilities when we do the actual implementation. The prototyping will help clarify the edge cases when the information about capabilities is necessary.

PR #34, opened 2021-11-17
Add explanation of agent_attributes field

tigrannajaryan:
The explanation was missing in the spec and it is important
to have recommendations about what exactly should go in the
attributes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -552,11 +552,9 @@ https://github.com/open-telemetry/opentelemetry-specification/tree/main/specific
 
 The following attributes SHOULD be included:
 - os.type, os.version - to describe where the agent runs.
-
-The following attributes are recommended to be included:
+- any other os.* attributes known.
 - host.* to describe the host the agent runs on.
 - cloud.* to describe the cloud where the host is located.
-- other os.* attributes not listed in the previous paragraph.
 - any other relevant Resource attributes that describe this agent and the
   environment it runs in.
 - any user-defined attributes that the end user would like to associate with
--- a/README.md
+++ b/README.md
@@ -544,6 +544,25 @@ The version number of the agent build. The Server can use this information for
 example to decide if it wants to offer a package of a different version to the
 agent via AgentPackageAvailable message.
 
+#### agent_attributes
+
+Attributes that describe the agent and the environment it runs in. Keys/values
+are according to OpenTelemetry semantic conventions, see:
+https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/resource/semantic_conventions
+
+The following attributes SHOULD be included:
+- os.type, os.version - to describe where the agent runs.
+
+The following attributes are recommended to be included:
+- host.* to describe the host the agent runs on.
+- cloud.* to describe the cloud where the host is located.
+- other os.* attributes not listed in the previous paragraph.
+- any other relevant Resource attributes that describe this agent and the
+  environment it runs in.
+- any user-defined attributes that the end user would like to associate with
+  this agent.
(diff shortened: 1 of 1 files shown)


djaglowski (approved):

README.md
```diff
@@ -544,6 +544,25 @@ The version number of the agent build. The Server can use this information for
...
+#### agent_attributes
+
+Attributes that describe the agent and the environment it runs in. Keys/values
+are according to OpenTelemetry semantic conventions, see:
+https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/resource/semantic_conventions
+
+The following attributes SHOULD be included:
+- os.type, os.version - to describe where the agent runs.
+
+The following attributes are recommended to be included:
```
Aneurysm9, line 556: What's the difference between `SHOULD` and `recommended`?  Since `SHOULD` is not `MUST` isn't it a recommendation instead of a requirement?

tigrannajaryan: I guess what I tried to express is that host.*, cloud.*, etc may not be known to the Agent. I probably should rephrase and say "The following attributes SHOULD to be included if they are available" or perhaps even that is unnecessary since "if available" is implied anyway.

tigrannajaryan: Removed the distinction, now it is just one list with a SHOULD keyword.

README.md
```diff
@@ -544,6 +544,23 @@ The version number of the agent build. The Server can use this information for
...
 
+#### agent_attributes
+
+Attributes that describe the agent and the environment it runs in. Keys/values
+are according to OpenTelemetry semantic conventions, see:
+https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/resource/semantic_conventions
+
+The following attributes SHOULD be included:
+- os.type, os.version - to describe where the agent runs.
+- any other os.* attributes known.
```
tigrannajaryan, line 555: These doesn't seem necessary. I think SHOULD is unnecessary for these.

tigrannajaryan: Deleted this.

README.md
```diff
@@ -544,6 +544,24 @@ The version number of the agent build. The Server can use this information for
...
+The following attributes SHOULD be included:
+- os.type, os.version - to describe where the agent runs.
+- host.* to describe the host the agent runs on.
+- cloud.* to describe the cloud where the host is located.
+- any other relevant Resource attributes that describe this agent and the
+  environment it runs in.
+- any user-defined attributes that the end user would like to associate with
+  this agent.
+
+Other attributes MAY be included to describe the Agent.
```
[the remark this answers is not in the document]
tigrannajaryan: Please see https://github.com/open-telemetry/opamp-spec#own-telemetry-reporting
>service.instance.id attribute SHOULD be set to Agent's instance_uid that is used in the OpAMP messages.

So, the `service.instance.id` is already reported via a mandatory `instance_uid` field in the AgentToServer message. Sending `service.instance.id` in the `agent_attributes` would be a duplicate of that.

Do we need that duplicate? I think it is not necessary. Perhaps we should make clear in this section why it is superfluous and how `instance_uid` field serves that need.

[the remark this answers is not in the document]
tigrannajaryan: Similarly for service name we have this:
>service.name attribute SHOULD be set to Agent's type, matching the value of agent_type field in AgentDescription message.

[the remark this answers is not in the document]
pmm-sumo: Agree, I missed that `AgentToServer` always includes `instance_uid`

pmm-sumo (approved):

Issue #31, opened 2021-11-16
Do we need addon versioning and support for multiple versions to be in use?


portertech:
Addons are currently defined as a file or set of files (related issue https://github.com/open-telemetry/opamp-spec/issues/5) with a download URL and a hash. If a different (or "newer") addon hash becomes available, the appropriate agents will download it and replace the existing addon. Should agents keep multiple addon versions/hashes? Should an addon have a specified version?

Use cases:
- Agent downloads a different addon hash, the addon errors, the agent is able to use the previous addon hash (or version)
- Agent uses multiple addon versions/hashes concurrently, using capabilities that may have been removed and added across addon versions/hashes

tigrannajaryan:
We have an [open issue](https://github.com/open-telemetry/opamp-spec/issues/8) about lack of "version" field for addons, but that issue does not describe any use cases. I am going to close that and keep this one since this actually brings good arguments about why it is necessary.

My +1 to adding the "version" field. I think we should do it.

tigrannajaryan:
This is resolved by https://github.com/open-telemetry/opamp-spec/pull/77 which unified addons and packages and the addons can now have versions.

PR #41, opened 2021-11-25
Add version field to Addons

tigrannajaryan:
Contributes to https://github.com/open-telemetry/opamp-spec/issues/31

The Addon version is added to the Server offers and to the Agent Status reports.

This covers the use case described in the issue:
"Agent downloads a different addon hash/version, the addon errors, the agent is
 able to use the previous addon hash/version".

This may or may not cover the second use case:
"Agent uses multiple addon hashes/versions concurrently, using capabilities that
 may have been removed and added across addon hashes/version".
The protocol does not allow reporting status of multiple versions concurrently,
but the Agent can still keep multiple versions around. It is not completely
clear from the description of the use case if this is sufficient.

### DIFF
--- a/README.md
+++ b/README.md
@@ -812,15 +812,17 @@ The AgentAddonStatus has the following structure:
 ```protobuf
 message AgentAddonStatus {
     string name = 1;
-    bytes agent_has_hash = 2;
-    bytes server_offered_hash = 3;
+    string agent_has_version = 2;
+    bytes agent_has_hash = 3;
+    string server_offered_version = 4;
+    bytes server_offered_hash = 5;
     enum Status {
         INSTALLED = 0;
         INSTALLING = 1;
         INSTALL_FAILED = 2;
     }
-    Status status = 4;
-    string error_message = 5;
+    Status status = 6;
+    string error_message = 7;
 }
 ```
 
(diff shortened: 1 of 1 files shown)


tigrannajaryan:
@portertech please review.

tigrannajaryan:
@open-telemetry/specs-approvers please review.

README.md
```diff
@@ -831,8 +833,17 @@ message AgentAddonStatus {
 Addon name. MUST be always set and MUST match the key in the addons field of
 AgentAddonStatuses message.
 
-<h4 id="agent_has_hash">agent_has_hash</h4>
+#### agent_has_version
 
+The version of the addon that the Agent has.
```
pmm-sumo, line 838: IIRC we discussed possibility for Agent having several addon versions installed locally (e.g. when the new one fails to be able to do a rollback). Should we add a clarification that when the agent has multiple addon versions, the most recent version should be reported (or perhaps the one that is enabled?)

tigrannajaryan: Can we leave the decision to the Agent and leave unspecified in the protocol?

pmm-sumo: Yeah, I think that should be OK for now and if we ever need to change that, we could revisit

pmm-sumo (approved):

Issue #16, opened 2021-11-09
Decide if the protocol can be used to manage SDKs


tigrannajaryan:
Can we use the same protocol to manage Otel SDKs?

Currently applicable parts:
- Remote Configuration Management (yes)
- Status Updates (unclear)
- Addons (no)
- Connection Settings Managements (maybe)
- Addons (no)
- Agent package update (no)

tigrannajaryan:
One thing that we will need to change if we want to manage SDK is the following current spec requirements:
>The Resource in the reported telemetry SHOULD describe the Agent in the following way:
>- service.instance.id attribute SHOULD be set to Agent's instance_uid that is used in the OpAMP messages.
>- service.name attribute SHOULD be set to Agent's type, matching the value of agent_type field in AgentDescription message.
>- service.version SHOULD be set to Agent's version, matching the value of agent_version field in AgentDescription message.

These are written with the assumption that the Agent is a standalone executable and we can call it a "Service". 

However this is not true for embedded agents, such as Otel SDK. Calling an Otel SDK a "Service" and requiring it to output its own metrics with "service.*" attributes would be very confusing. For an Otel SDK a more natural way would be to emit its own metrics for example with the following attributes:
- telemetry.sdk.name (this already exists as Otel semantic convention).
- telemetry.sdk.version (we can add this as a new Otel semantic convention).
- telemetry.sdk.instance.id (we can add this as a new Otel semantic convention and require it to be globally unique, so technically this one alone is sufficient as an identifying attribute).

To make such usage possible with OpAMP we need to modify OpAMP spec in the following way (this is just one possible way, there are probably others):
- In AgentDescription message record identifying and non-identifying attributes separately. For example for a standalone Agent the following would be identifying attributes: service.instance.id, service.name, service.version, while e.g. os.type would be non-identifying.
- For Own Telemetry Reporting instead of requiring the hard-coded list of attributes (service.instance.id, service.name, service.version), instead require that the identifying attributes that are included in the AgentDescription message MUST be also used as attributes in own telemetry.
This gives the flexibility for the agents to use any attributes they want to identify themselves, yet still clearly establishes the link between OpAMP's Agent and its own telemetry collected elsewhere.

@open-telemetry/opamp-spec-approvers what do you think?

tigrannajaryan:
@MrAlias @jmacd what do you think about the above comment? Would it make sense for Otel SDK to follow the conventions like that and use OpAMP to receive the config?

[...]

PR #48, opened 2021-12-03
Generalize Agent identification

tigrannajaryan:
Resolves https://github.com/open-telemetry/opamp-spec/issues/16

This change allows the protocol to be used with non-standalone
Agents, such as OpenTelemetry SDK. For such Agents the identification
is not done via `service.*` attributes, but via a different set of
attributes (e.g. `telemetry.sdk.*` attributes).

This change replaces (agent_type,agent_version) tuple by a list of
attributes which can be freely chosen by the Agent. The list is
further split into 2 parts: identifying, which is required for linking
to own telemetry, and non-identifying which allow additional description.

We also need to submit a PR against Otel spec repo to add uniquely
identifying attributes for Otel SDKs (e.g. telemetry.sdk.instance.id).

### DIFF
--- a/README.md
+++ b/README.md
@@ -632,39 +632,47 @@ enum AgentCapabilities {
 
 <h3 id="agentdescription-message">AgentDescription Message</h3>
 
-
 The AgentDescription message has the following structure:
-
 
 ```protobuf
 message AgentDescription {
-    string agent_type = 1;
-    string agent_version = 3;
-    repeated KeyValue agent_attributes = 4;
+    repeated KeyValue identifying_attributes = 1;
+    repeated KeyValue non_identifying_attributes = 2;
 }
 ```
 
+#### identifying_attributes
 
-<h4 id="agent_type">agent_type</h4>
+Attributes that identify the agent.
 
-
-The reverse FQDN that uniquely identifies the agent type, e.g.
-"io.opentelemetry.collector".
-
-<h4 id="agent_version">agent_version</h4>
-
-
-The version number of the agent build. The Server can use this information for
-example to decide if it wants to offer a package of a different version to the
-agent via AgentPackageAvailable message.
-
-#### agent_attributes
-
-Attributes that describe the agent and the environment it runs in. Keys/values
-are according to OpenTelemetry semantic conventions, see:
+Keys/values are according to OpenTelemetry semantic conventions, see:
 https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/resource/semantic_conventions
 
+For standalone running Agents (such as OpenTelemetry Collector) the following
+attributes SHOULD be specified:
+
+- service.name should be set to a reverse FQDN that uniquely identifies the
+  agent type, e.g. "io.opentelemetry.collector"
+- service.namespace if it is used in the environment where the Agent runs.
+- service.version should be set to version number of the Agent build.
+- service.instance.id should be set. It may be be set equal to the Agent's
+  instance uid (equal to ServerToAgent.instance_uid field) or any other value
+  that uniquely identifies the Agent in combination with other attributes.
+- any other attributes that are necessary for uniquely identifying the Agent's
+  own telemetry.
+
+The Agent SHOULD also include these attributes in the Resource of its own
+telemetry. The combination of identifying attributes SHOULD be sufficient to
+uniquely identify the Agent's own telemetry in the destination system to which
+the Agent sends its own telemetry.
+
+#### non_identifying_attributes
+
+Attributes that do not necessarily identify the Agent but help describe where it
+runs.
+
 The following attributes SHOULD be included:
+
 - os.type, os.version - to describe where the agent runs.
 - host.* to describe the host the agent runs on.
 - cloud.* to describe the cloud where the host is located.
@@ -672,9 +680,6 @@ The following attributes SHOULD be included:
   environment it runs in.
 - any user-defined attributes that the end user would like to associate with
   this agent.
-
-Other attributes MAY be included to describe the Agent.
-
 
 <h3 id="effectiveconfig-message">EffectiveConfig Message</h3>
 
@@ -1213,14 +1218,12 @@ message ConnectionSettingsOffers {
 ```
 
 
-<h4 id="hash">hash</h4>
-
+#### hash
 
 Hash of all settings, including settings that may be omitted from this message
 because they are unchanged.
 
-<h4 id="opamp">opamp</h4>
-
+#### opamp
 
 Settings to connect to the OpAMP server. If this field is not set then the agent
 should assume that the settings are unchanged and should continue using existing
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -632,37 +632,42 @@ enum AgentCapabilities {
...
-"io.opentelemetry.collector".
+Attributes that identify the agent. Keys/values are according to OpenTelemetry
+semantic conventions, see:
+https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/resource/semantic_conventions
 
-<h4 id="agent_version">agent_version</h4>
+For standalone running Agents (such as OpenTelemetry Collector) the following
+attributes SHOULD be specified:
 
+- service.name should be set to a reverse FQDN that uniquely identifies the
```
pmm-sumo, line 653: Let's say that the agent in question is OpenTelemetry Java Instrumentation. How can we inform the server about this? 

[service.name](https://github.com/open-telemetry/opentelemetry-specification/blob/da74730bf835010229a9612c281f052e26553218/specification/resource/semantic_conventions/README.md#service) could be already set to something like `shoppingcart` and I am thinking that using it for `io.opentelemetry.collector` might not be a good fit here. 

Perhaps `service.name` could be optional and we should have a new convention (`service.type`?). It could hold values such as `opentelemetry-collector`, `fluentbit`, `opentelemetry/java-auto-instrumentation` or such

tigrannajaryan: Java Instrumentation will not use `service.name` to describe itself. `service.*` attributes are intended to describe the Application (or the Service) that is being instrumented, not the SDK or the library that does the instrumentation.

Java instrumentation is part of an Otel SDK and Otel SDK uses a different set of attributes to describe itself. We already have some [conventions](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/resource/semantic_conventions#telemetry-sdk) for the SDKs. Java Instrumentation will set:
telemetry.sdk.name="opentelemetry"
telemetry.sdk.language="java"
telemetry.sdk.version="some version number" (and/or `telemetry.auto.version`).

These are probably not sufficient though. We will likely also want to have a unique identifier that identifies the particular running instance of the application instrumented with Java SDK. This is important if we want SDKs or instrumentation libraries to report their own telemetry (e.g. send metrics about the number of telemetry items they collected). For this I think we can add a convention `telemetry.sdk.instance.id` (by analogy of `service.instance.id`). SDKs will autogenerate the instance id at startup or will allow to specify it via an env variable if persistent id is wanted. Or perhaps the SDK could also set `telemetry.sdk.instance.id` equal to `service.instance.id` if the latter is provided but the former is not provided.

pmm-sumo: My point is that `service.name` here is said to "uniquely identify the agent type, e.g. 'io.opentelemetry.collector'" which I think is not true. `service.name` has different purpose and we might want to have some other field to identify the type. 

Perhaps this could be done via checking fields like `telemetry.sdk.*` (for SDK) but still it does not say how to detect that the agent is, say, OpenTelemetry Collector or Fluentbit. We had `agent_type` field previously, which is now gone

pmm-sumo: So `service.name=io.opentelemetry.collector` for OpenTelemetry Collector, `service.name=io.fluentbit` for Fluentbit and so forth. If the value is unknown (e.g. `service.name=foo`), it indicates it's an SDK and one needs to look to `telemetry.sdk.*` to understand what kind of SDK is being employed?

tigrannajaryan: >So service.name=io.opentelemetry.collector for OpenTelemetry Collector, service.name=io.fluentbit for Fluentbit and so forth.

Yes.

>If the value is unknown (e.g. service.name=foo), it indicates it's an SDK and one needs to look to telemetry.sdk.* to understand what kind of SDK is being employed?

No. If `service.name` but the value is unknown then it is still valid and the server should assume it is a standalone Agent, just the kind of an Agent that is unknown to the Server (what Server does in that case is Server's business and is outside of the protocol spec).

The Server generally doesn't need to look for any specific attribute (`service.*` or `telemetry.sdk.*`). It should just:
1. Use what is listed in `identifying_attributes` as a set of attributes to link to the Agent's own telemetry sent elsewhere.
2. Use a `identifying_attributes` and `non_identifying_attributes` as a set of attributes over which the Server may define any business logic / rules to return appropriate configuration to the Agent.

The Server doesn't necessarily need to have any hard-coded logic for specific values of `service.name`, although it can have such logic if wanted. The Server may for example have a specific hard-coded logic for the case service.name="io.opentelemetry.collector" and another hard-coded for for the case when telemetry.sdk.name="opentelemetry". Again, these are implementation details of the Server, they are not protocol concerns.

Real-world Server implementations will likely employ a combination of hard-coded rules (for specific attributes values), generic rules (e.g. for linking to own telemetry of arbitrary agents without know what agent it is) and user-defined rules (e.g. to allow users to override config for specific agents or subsets of agents).

carlosalberto:
> Java Instrumentation will not use service.name to describe itself. service.* attributes are intended to describe the Application (or the Service) that is being instrumented, not the SDK or the library that does the instrumentation.

> Java instrumentation is part of an Otel SDK and Otel SDK uses a different set of attributes to describe itself. We already have some conventions for the SDKs. Java Instrumentation will set:
telemetry.sdk.name="opentelemetry"
telemetry.sdk.language="java"
telemetry.sdk.version="some version number" (and/or telemetry.auto.version).

This is a good explanation of why SDKs are not supposed to be setting `service.*` Resource attributes. Consider adding it to the changes themselves?

Aneurysm9:
I think this is a good change.  Allowing the initiator to determine which attributes are relevant for identification seems better than having to shoehorn in values that don't necessarily match the concepts represented by the pre-defined attributes.  This is similar to the discussions recently had over `service.name` vs. `application.name`.  Avoiding that issue before this spec is stabilized would be good.

pmm-sumo (approved):

pmm-sumo:
One related thing that I am wondering about - and maybe I missed something - is how accurately the `identifying_attributes` can identify the specific agent installation (rather than runtime instance)?

I.e. my understanding is that attributes such as `service.instance.id` are changing each time given instance restarts, event if this is the same host.

Should we have some additional means to identify the given agent installation (for a lack of a better term)? This would help with tracking events related to a specific place where agent installed. I believe we could use `host.id` or such for that purpose (which could be part of identifying attributes), though wanted to raise it to make sure we're not missing something

tigrannajaryan:
> I.e. my understanding is that attributes such as `service.instance.id` are changing each time given instance restarts, event if this is the same host.

I think we should recommend the Agents who have access to persistent storage to persist the id and reuse it. This can allow the Server to store a specific config for the particular instance and return it. If the id is not persisted then you are right it is only useful for the duration of the process uptime and other attributes need to be used for identification.

Perhaps we need to make this clearer in the definition of what "identifying" means. For now it refers to the identification of own telemetry in the telemetry backend. The telemetry backend should have the same or stronger uniqueness requirements as the OpAMP Server. This means that identifying_attributes should be definitely able to uniquely identify a specific agent installation (otherwise reported own telemetry will be mixed up in the telemetry backend).

pmm-sumo:
> > I.e. my understanding is that attributes such as `service.instance.id` are changing each time given instance restarts, event if this is the same host.
> 
> I think we should recommend the Agents who have access to persistent storage to persist the id and reuse it. 

Should we open an issue for OpenTelemetry Collector that, if possible, the instance id should be persisted? (This would have another benefit of limiting time-series cardinality for own telemetry)

tigrannajaryan:
> > > I.e. my understanding is that attributes such as `service.instance.id` are changing each time given instance restarts, event if this is the same host.
> > 
> > 
> > I think we should recommend the Agents who have access to persistent storage to persist the id and reuse it.
> 
> Should we open an issue for OpenTelemetry Collector that, if possible, the instance id should be persisted? (This would have another benefit of limiting time-series cardinality for own telemetry)

I was thinking that the instance id will be generated by the Supervisor and will be persisted by Supervisor. The Supervisor will put the instance id in the Collector's config file (generated by Supervisor). For this to work we will need to add support to specify the "service.instance.id" to the `telemetry` section of Collector config file.

tigrannajaryan:
This is now ready for review. Once accepted I will go ahead with the corresponding implementation in https://github.com/open-telemetry/opamp-go/pull/43

tigrannajaryan:
@open-telemetry/opamp-spec-approvers please review.

djaglowski (approved):

PR #55, opened 2022-01-13
Donate OpAMP specification to OpenTelemetry

tigrannajaryan:
After this PR is merged to mark the formal donation I will submit a separate PR that merges the content with the modifications that we did earlier.

This donation is done according to https://github.com/open-telemetry/community/blob/main/CONTRIBUTING.md#donations
and falls into the category of small donations and as such does not require the GC and TC process that the large donations follow.


pmm-sumo (approved):

xxkap0wxx (approved):
Approving on behalf of Splunk OSS committee

Issue #56, opened 2022-01-13
Agent registration process


pmm-sumo:
Currently, OpAMP Client can practically identify itself using two means:
* `instance_uid` field, included in AgentToServer message
* via HTTP authentication fields (using either Basic Auth or Bearer) - if it is using unique credentials

The UID generation is a concern of the agent, which might limit some use cases where certain formats of UID can be used to identify different groups of agents and as such should be provided by server, also ensuring that two agent instances always use different ID's.

For HTTP authentication, there are different use-cases and sometimes a single set of credentials can be used for a larger group of clients. However, it might be preferred to have an initial, installation-only auth token that is used only to identify the type of the agent and handle the registration after which the agent is provided with a unique token that will be used for any further communication. The unique identifier of the agent can be embedded into such token.

Perhaps there's a room to provide support for agent registration procedure, in which a registration-only token is used to get the agent-unique token that will be persisted and used for future communication?

mtwo:
Following up from the weekly call, we should discuss this next session once @pmm-sumo is back. We'd like to understand more about the scenarios that require the unique ID to be provided by the server.

tigrannajaryan:
Discussed in Workgroup meeting. We can add ability for the Server to command the Agent to use a different UID. We can also make UUID optional in the first message from the Agent. The combination of these 2 features allows to:
- Agent to request the Server to assign it a UUID on first connect.
- Server to detect duplicate UUIDs and order Agents to use a different UUID.

pmm-sumo:
I promised to prepare a PR with proposal of the spec update, got a bit delayed but should have it this week

PR #63, opened 2022-02-22
Set instance_uid by Server on conflict or request for generation

pmm-sumo:
This is a first approach towards #56 

It assumes that Agent's `instance_uid` can always be updated by Server (since resolving potential issues with conflicts is important).

There's a gap with multiplexed websocket connections which perhaps could be handled by adding Agent capability flag indicating if overrides can be accepted or not


tigrannajaryan:
> There's a gap with multiplexed websocket connections 

Yeah, this is unfortunate. To avoid this problem perhaps we can do slightly differently. Instead of using empty `instance_uid` field in AgentToServer message as an indication of a request to assign a new instance_uid we can have an explicit flag in AgentToServer to indicate it. In that case the `instance_uid` field in AgentToServer message may still be set to some temporary value by the Agent until it gets the new instance_uid value from the Server.

Another approach is for the terminating proxy to NOT multiplex connections with empty incoming `instance_uid` field with other connections. So the initial assignment of instance_uid will require a dedicated outgoing connection from the proxy to the Server. After the assignment of the instance_uid the dedicated connection can be dropped and the multiplexing can start.

>which perhaps could be handled by adding Agent capability flag indicating if overrides can be accepted or not

I am not sure how this will solve the problem. Can you expand?

specification.md
```diff
@@ -318,6 +329,9 @@ enum Flags {
     // set if the Agent indicated it cannot report addon status by setting
     // the ReportsAddonStatus bit to 0 in StatusReport.capabilities field.
     ReportAddonStatus     = 0x00000002;
+
+    // The server requests the agent to change its instance_uid to the provided value.
+    UpdateInstanceUid     = 0x00000003;
```
tigrannajaryan, line 334:
```suggestion
    UpdateInstanceUid     = 0x00000004;
```
Flags is a bit mask.

pmm-sumo:
> > There's a gap with multiplexed websocket connections
> 
> Yeah, this is unfortunate. To avoid this problem perhaps we can do slightly differently. Instead of using empty `instance_uid` field in AgentToServer message as an indication of a request to assign a new instance_uid we can have an explicit flag in AgentToServer to indicate it. In that case the `instance_uid` field in AgentToServer message may still be set to some temporary value by the Agent until it gets the new instance_uid value from the Server.

I thought about it and I am leaning towards that approach. Will update the PR

> > which perhaps could be handled by adding Agent capability flag indicating if overrides can be accepted or not
> 
> I am not sure how this will solve the problem. Can you expand?

This would merely inform that it is not possible to override the agent instance_id. I think the idea proposed above (temporary id's and explicit flag) is a better one

specification.md
```diff
@@ -159,6 +159,7 @@ message AgentToServer {
     AgentAddonStatuses addon_statuses = 3;
     AgentInstallStatus agent_install_status = 4;
     AgentDisconnect agent_disconnect = 5;
+    AgentToServerFlags flags = 6;
```
pmm-sumo, line 162: ServerToAgent has "Flags", so I used "AgentToServerFlags" here but maybe both of them could use a better names. What do you think @tigrannajaryan ?

specification.md
```diff
@@ -1678,9 +1706,14 @@ The message has the following structure:
 ```protobuf
 message AgentRemoteConfig {
   AgentConfigMap config = 1;
+  string new_instance_uid = 2;
```
pmm-sumo, line 1709: I wasn't sure about the right place for this field. Did not want to create a new message and concluded that in fact it's part of the config, but I am open to suggestions on moving it elsewhere. Since we want to support multiplexing when setting instance_uid, it needs to be separated from the "ServerToAgent.instance_uid"

specification.md
tigrannajaryan: AgentToServerFlags sounds good to me. If you can think of a better name please suggest.

specification.md
tigrannajaryan: I do not think it fits well in the AgentRemoteConfig. The "config" is defined to mean a set of opaque byte sequences to the protocol. This now violates that notion.
A new message sounds better to me.

pmm-sumo: Sure, added `AgentIdentification` as a new field

specification.md
```diff
@@ -357,6 +383,17 @@ enum ServerCapabilities {
 }
 ```
 
+#### agent_identification
```
pmm-sumo, line 386: I think that "new_instance_uid" would be too specific field and wanted something with a bit broader scope so I thought about "AgentIdentification", though also wondering about "AgentIdentifcationOverride", "AgentIdentificationOffer" or so. If you had anything else in mind @tigrannajaryan I will gladly accept it

tigrannajaryan: Is there anything else that we may want to add to AgentIdentification message in the future?

pmm-sumo: > Is there anything else that we may want to add to AgentIdentification message in the future?

I don't have anything on my mind right now though I'm actually wondering if instead creating `AgentIdentification`, a model used in `AgentDescription` could be followed (so repeated KeyValue). I don't really have a strong use case but wondering if this would give an opportunity for the server to send some attributes back (both identifying and non-identifying) that would be persisted by the agent. Not sure if it makes much sense though (since server could always attach them later itself if needed)

tigrannajaryan: We can add keyvalue list as an additional field later if needed. I think what we have for now is fine.

tigrannajaryan (approved):

pmm-sumo:
> @pmm-sumo will you be able to submit a corresponding change in opamp-go client and server implementations?

@tigrannajaryan sure thing, will have it later today

Issue #67, opened 2022-02-28
content_hash of Downloadable file needs to define hashing method


tigrannajaryan:
Without specifying the hashing method it is a useless for the purposes it claims to exist for
>The hash of the file content. Can be used by the Agent to verify that the file was downloaded correctly.

You can't verify that the hash matches the content if you don't know the hashing method.

PR #68, opened 2022-03-02
Define that DownloadFile content hash method is SHA256

tigrannajaryan:
Resolves https://github.com/open-telemetry/opamp-spec/issues/67

Without specifying the hashing method it was useless for the purposes it claims to exist for
>The hash of the file content. Can be used by the Agent to verify that the file was downloaded correctly.

You can't verify that the hash matches the content if you don't know the hashing method.

Now it is precisely defined to be SHA256 of the content.

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -1851,7 +1851,7 @@ There are 3 levels of hashes:
 
 <h4 id="file-hash">File Hash</h4>
 
-The hash of the addon file content. This is stored in the [hash](#hash) field in
+The hash of the addon file content. This is stored in the [content_hash](#content_hash) field in
 the [DownloadableFile](#downloadablefile-message) message. This value SHOULD be
 used by the Agent to determine if the particular file it has is different on the
 Server and needs to be re-downloaded.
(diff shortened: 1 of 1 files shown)


andykellr:
This seems reasonable, but I had assumed that the exact hashing algorithm was shared between the client and server, similar to your comment in #69 that, "The exact signing and verification method is Agent specific."

pmm-sumo:
As discussed during the meeting, I think it might be more flexible to have name of the hashing method provided separately rather than fixing on SHA-256. So e.g. `string hashing_algorithm = 3;`

pmm-sumo:
As discussed during the meeting, I think it might be more flexible to have name of the hashing method provided separately rather than fixing on SHA-256. So e.g. `string hashing_algorithm = 3;`

tigrannajaryan:
I think the idea that we may need other hash methods is valid. I would not want to add support for other methods right now, but we need to make it possible in the future. To enable this I will rename `content_sha256` to `content_hash`, but will keep the notice that it must be SHA256. In the future we can add `hashing_algorithm` and make "sha256" the default value for it, so it will be a backward compatible addition. I think this is a good compromise to avoid adding too much to the spec right now.

pmm-sumo (approved):

pmm-sumo:
> @pmm-sumo PTAL, I made changes as described [above](https://github.com/open-telemetry/opamp-spec/pull/68#issuecomment-1063271283).

Sorry for the dealy, I must have missed the update. I think it's a fair approach

Issue #50, opened 2021-12-13
OpAMP over plain HTTP transport


tigrannajaryan:
Certain applications may not need the OpAMP's capabilities enabled by WebSocket (ability to push configs from server without Agent's polling) and may prefer to implement a simpler HTTP transport. This is likely the case for short-lived Agents such as for example the ones running in AWS lambda. The HTTP transport is likely simpler to implement on the Server (easier to manage connections) and also removes an additional dependency (a WebSocket library) from the client.

We are also considering implementing OpAMP in OpenTelemetry SDKs, which can result in significantly larger number of OpAMP Agents connecting to the Server, so the Server has to handle a larger number of persistent WebSocket connections, which may pose more burden than necessary for use cases where there are no benefits from having a persistent connection (e.g. the above mentioned short-lived use case).

Would it be useful to define a plain HTTP transport for OpAMP? It is fairly simple to modify the spec to allow it, with AgentToServer/ServerToAgent messages exchanged via HTTP POST Body.

It is also not very complicated to support both WebSocket and plain HTTP on the same Server and if we want to make it seamless from Agent's perspective we may need to add an indicator of the desired transport in the initial connection attempt made by the Agent, so that the Server knows whether to upgrade to WebSocket or just reply in plain HTTP.

carlosalberto:
I cannot speak much about the Collector case, but for SDKs using HTTP instead of websockets is definitely a good thing.

pmm-sumo:
I think it would be good to have such option, especially since it fits the communication model. Less dependencies for SDK's is a good thing too

tigrannajaryan:
Update: I quickly put some code together and it took about 100 lines to support plain HTTP in our server implementation: https://github.com/open-telemetry/opamp-go/compare/main...tigrannajaryan:plain-http-transport

The client-side plain HTTP implementation will require more work and the differences will be more substantial, but I feel good that a single server can simultaneously support both WebSocket and plain HTTP OpAMP connections on a single port with little extra complexity over the WebSocket-only implementation.

yurishkuro:
I may be missing it, but where is OpAMP defined? I only see Go reference implementation repos.

PR #70, opened 2022-03-22
Add plain HTTP transport to OpAMP

tigrannajaryan:
Resolves https://github.com/open-telemetry/opamp-spec/issues/50

Certain applications may not need the OpAMP's capabilities enabled by WebSocket (ability to push configs from server without Agent's polling) and may prefer to implement a simpler HTTP transport. This is likely the case for short-lived Agents such as for example the ones running in AWS lambda. The HTTP transport is likely simpler to implement on the Server (easier to manage connections) and also removes an additional dependency (a WebSocket library) from the client.

We are also considering implementing OpAMP in OpenTelemetry SDKs, which can result in significantly larger number of OpAMP Agents connecting to the Server, so the Server has to handle a larger number of persistent WebSocket connections, which may pose more burden than necessary for use cases where there are no benefits from having a persistent connection (e.g. the above mentioned short-lived use case).

This changes defines a plain HTTP transport for OpAMP. It is a fairly simple modification of the spec to allow it, with AgentToServer/ServerToAgent messages exchanged via HTTP POST Body.

It is also not very complicated to support both WebSocket and plain HTTP on the same Server and if we want to make it seamless from Agent's perspective we can add an indicator of the desired transport in the initial connection attempt made by the Agent, so that the Server knows whether to upgrade to WebSocket or just reply in plain HTTP. The "Content-Type" request header can be such indicator. We can add this in a follow-up PR.

Here is a draft Go OpAMP server implementation for reference: https://github.com/open-telemetry/opamp-go/compare/main...tigrannajaryan:plain-http-transport

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -84,10 +84,28 @@ for OpAMP Agents and Servers. The OTLP/HTTP protocol is
 The protocols used by the Agent to connect to other destinations are agent
 type-specific and are outside the scope of this specification.
 
-OpAMP protocol uses [WebSocket](https://datatracker.ietf.org/doc/html/rfc6455)
-as the transport. The Agent is a WebSocket client and the Server is a WebSocket
-server. The Agent and the Server communicate using binary data WebSocket
-messages. The payload of each WebSocket message is a
+OpAMP protocol works over one of the 2 supported transports: plain HTTP
+connections and WebSocket connections. Server implementations SHOULD accept both
+plain HTTP connections and WebSocket connections. Client implementations may
+choose to support either plain HTTP or WebSocket transport, depending on their
+needs.
+
+Typically a single Server accepts connections from many agents. Agents
+are identified by self-assigned globally unique instance identifiers (or
+instance_uid for short). The instance_uid is recorded in each message sent from
+the Agent to the Server and from the Server to the Agent.
+
+The default URL path for the connection is /v1/opamp. The URL path MAY be
+configurable on the Agent and on the Server.
+
+
+## WebSocket Transport
+
+One of the supported transports for OpAMP protocol is
+[WebSocket](https://datatracker.ietf.org/doc/html/rfc6455). The Agent is a
+WebSocket client and the Server is a WebSocket server. The Agent and the Server
+communicate using binary data WebSocket messages. The payload of each WebSocket
+message is a
 [binary serialized Protobuf](https://developers.google.com/protocol-buffers/docs/encoding)
 message. The Agent sends AgentToServer Protobuf messages and the Server sends
 ServerToAgent Protobuf messages:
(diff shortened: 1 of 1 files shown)


anuraggujral:
In the above PR, it is not clear from the second paragraph that here we are considering OpenTelemetry SDKs as agents, may be very should make it more explicit.

tigrannajaryan:
@pmm-sumo what do you think about this change?

specification.md
```diff
@@ -145,10 +163,47 @@ configuration.
...
+The WebSocket transport is typically used when it is necessary to have instant
+communication ability from the Server to the Agent without waiting for the Agent
+to poll the Server like it is done when using the HTTP transport (see below).
+
+## Plain HTTP Transport
+
+The second supported transport for OpAMP protocol is plain HTTP connection. The
+Agent is an HTTP client and the Server is an HTTP client server. The Agent makes
+POST requests to the Server. The body of the POST request and response is a
+[binary serialized Protobuf](https://developers.google.com/protocol-buffers/docs/encoding)
```
pmm-sumo, line 175: I think it would be worth considering allowing JSON serialized Protobuf as well. It typically has less dependencies and in some cases (e.g. mobile apps, web browsers, Node.js) is preferred over Protobuf

specification.md
```diff
@@ -145,10 +163,47 @@ configuration.
...
+the Server sends ServerToAgent Protobuf messages in the response body.
+
+OpAMP over HTTP is a synchronous, half-duplex message exchange protocol. The
+Agent initiates an HTTP request when it has an AgentToServer message to deliver.
+The Server responds to each HTTP request with a ServerToAgent message it wants
+to deliver to the Agent. If the Agent has nothing to deliver to the Server the
+Agent MUST periodically poll the Server by sending an AgentToServer message
+where only [instance_uid](#instance_uid) field is set. This gives the Server an
+opportunity to send back in the response any messages that the Server wants to
+deliver to the Agent (such as for example a new remote configuration).
```
pmm-sumo, line 186: I think this is the right approach. Polling also acts as a heartbeat from the agent

specification.md
```diff
@@ -145,10 +163,47 @@ configuration.
...
+
+The default polling interval when the Agent does have anything to deliver is 30
+seconds. This polling interval SHOULD be configurable on the Agent.
+
+When using HTTP transport the sequence messages is exactly the same as it is
+when using the WebSocket transport. The only difference is in the timing in the
+situation when the Server wants to send a message to the Agent but the Server
+needs to wait for the Agent to poll the Server and establish an HTTP request
+over which the Server's message can be sent back.
```
pmm-sumo, line 195: It took me a while to process the last sentence so just wanted to make sure I got it right. When Agent has something to send to the Server, it does not need to wait for the polling interval to kick, it can send the message anytime, for which the server might return another message (and the conversation might continue further). 

When Server has something to send to the Agent (e.g. ConnectioSettingsOffer), it might need to wait for the "ping" from the Agent.

specification.md
```diff
@@ -145,10 +163,47 @@ configuration.
...
+
+The Agent MUST set "Content-Type: application/x-protobuf" request header when
+using plain HTTP transport. When the Server receives an HTTP request with this
+header set it SHOULD assume this is a plain HTTP transport request, otherwise it
+SHOULD assume this is a WebSocket transport initiation.
+
```
pmm-sumo, line 201: Should we add something about compression or perhaps we just assume it's a transport propety?

pmm-sumo:
> @pmm-sumo what do you think about this change?

I think it will be a very useful addition. I think it might be worth running by OpenTelemetry SDK's maintainers, since this is one of the use cases for it

specification.md
tigrannajaryan: Yes, that's fairly easily achievable via Content-Type header and also easy to implement on the server-side to support both JSON and binary Protobufs.

specification.md
```diff
@@ -145,10 +163,47 @@ configuration.
...
+to deliver to the Agent. If the Agent has nothing to deliver to the Server the
+Agent MUST periodically poll the Server by sending an AgentToServer message
+where only [instance_uid](#instance_uid) field is set. This gives the Server an
+opportunity to send back in the response any messages that the Server wants to
+deliver to the Agent (such as for example a new remote configuration).
+
+The default polling interval when the Agent does have anything to deliver is 30
+seconds. This polling interval SHOULD be configurable on the Agent.
+
+When using HTTP transport the sequence messages is exactly the same as it is
```
tigrannajaryan, line 191:
```suggestion
When using HTTP transport the sequence of messages is exactly the same as it is
```

specification.md
tigrannajaryan: Correct. I probably need to rewrite this to make it clear.

specification.md
tigrannajaryan: Good question. I think for the response compression we can leave it to be negotiated via Accept-Encoding/Content-Encoding headers.

There is no standard HTTP way to do the negotiation for the request. So, if we want the compression possibility either we mandate its support in the spec or leave it undefined (and maybe say that it is user configurable). For OTLP we chose to mandate "none" and "gzip" in the spec: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#protocol-details
So maybe we should do the same here.

specification.md
tigrannajaryan: Added as Open Question.

specification.md
tigrannajaryan: Added to the text.

specification.md
tigrannajaryan: Made some clarifications.

pmm-sumo (approved):

Issue #71, opened 2022-03-28
Description of "Operation: Status Reporting by agent" uses outdated field body


anuraggujral:
In the StatusReporting Operation , it is mentioned that the status report is sent as an AgentToServer message, where the body field is set to StatusReport message. I thought the whole AgentToServer message will go as body and not just StatusReport. Filing this issue to correct existing documentation.

pmm-sumo:
Thanks for reporting this @anuraggujral - the fix has been just merged

PR #74, opened 2022-03-30
Remove usages of #Body

pmm-sumo:
Here's a take on fixing #71 

I looked for `#Body` references and removed all I could find

@tigrannajaryan I wasn't fully sure of wording. Did not want to change the structure too much, if you think it could be improved then any suggestions are more than welcome


tigrannajaryan (approved):
Thanks for the fix. LGTM.

Issue #4, opened 2021-11-05
Is it possible to unify addons and agent package such that we don't have 2 different concepts to deal wiith


tigrannajaryan:
The spec defines downloadable addons and downloadable agent package. These features overlap in some ways. Can we unify and get of one of the concepts?

tigrannajaryan:
We can do this by deleting the entire spec section dedicated to agent packages and then renaming "addon" to "package" in the spec and adding an enum to the messages to tell if the package is an addon package or agent's installable package. This should significantly simplify the protocol and remove duplication of very similar functionality.

PR #77, opened 2022-04-25
Unify addons and agent packages

tigrannajaryan:
Addons and agent packages were previously defined separately
in the spec but had a lot of similarities. This change combines
them into one concept of "packages". The resulting spec is
significantly shorter while providing the same functionality.
It may also reduce the code duplication in implementations.

Resolves https://github.com/open-telemetry/opamp-spec/issues/4

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -2159,10 +2159,6 @@ enum PackageType {
 }
 ```
 
-TODO: is there a need to prescribe any specific behavior or explain what happens if there
-are more than one top-level packages in the server's offer or we can leave it out and
-allow agent-specific behavior?
-
 #### version
 
 The package version that is available on the server side. The agent may for
--- a/specification.md
+++ b/specification.md
@@ -23,8 +23,7 @@ Note 2: this document requires a simplification pass to reduce the scope, size a
   * [AgentToServer Message](#agenttoserver-message)
       - [instance_uid](#instance_uid)
       - [status_report](#status_report)
-      - [addon_statuses](#addon_statuses)
-      - [agent_install_status](#agent_install_status)
+      - [package_statuses](#package_statuses)
       - [agent_disconnect](#agent_disconnect)
       - [flags](#flags)
   * [ServerToAgent Message](#servertoagent-message)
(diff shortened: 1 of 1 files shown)


tigrannajaryan:
@open-telemetry/opamp-spec-approvers I made this a draft for now, please tell what you think about it.

tigrannajaryan:
@andykellr it would be also good to have your thoughts on this.

andykellr:
I can look at it in more detail tomorrow or Wednesday, but at first glance it makes sense to me.

tigrannajaryan:
@open-telemetry/opamp-spec-approvers I spent a bit more time thinking about this change and feel reasonably confident that this is a move in the right direction I think this is ready for review.

specification.md
```diff
@@ -2153,110 +2064,121 @@ There are 3 levels of hashes:
...
+enum PackageType {
+    TopLevelPackage = 0;
+    AddonPackage    = 1;
+}
+```
+
+TODO: is there a need to prescribe any specific behavior or explain what happens if there
+are more than one top-level packages in the server's offer or we can leave it out and
+allow agent-specific behavior?
+
```
andykellr, line 2165: I think this can be agent specific.

Related to this, I think a `name string` could be useful on the PackageAvailable or DownloadableFile message identifying the package. This may not necessarily be determined from the download_url.

andykellr (approved):
I like this change.

specification.md
tigrannajaryan: >I think this can be agent specific.

I agree, I thought the same way. I will delete the TODO.

>Related to this, I think a name string could be useful on the PackageAvailable or DownloadableFile message identifying the package. This may not necessarily be determined from the download_url.

`PackageAvailable` is only referenced from `packages` map in `PackagesAvailable` field, where the keys of the map are the package names. Adding the `name` to `PackageAvailable` would be a duplicate. Or are you thinking about something else?

For `DownloadableFile`: are you looking for the package name or the file name here?

andykellr: That makes sense. I think the key in the map is fine.

tigrannajaryan:
@pmm-sumo it would be great if you could also have a look at this.

Issue #76, opened 2022-04-20
Decide if AgentDescription must be included in every StatusReport


tigrannajaryan:
See the comment here: https://github.com/open-telemetry/opamp-go/pull/63#issuecomment-1098526891

The spec says:
```
The Agent MUST send a status report:
* First time immediately after connecting to the Server. The status report MUST
  be the first message sent by the Agent.
* Subsequently every time the status of the Agent changes.
```

Nowhere does it says that the AgentDescription must be also included in the status report. On the contrary, [the spec says](https://github.com/open-telemetry/opamp-spec/blob/410dc86792d6109050e767a1d76de12a5b9f0129/specification.md?plain=1#L624):
```
<h4 id="agent_description">agent_description</h4>

The description of the agent, its type, where it runs, etc. See
[AgentDescription](#agentdescription-message) message for details.

This field SHOULD be unset if no Agent description fields have changed since the
last StatusReport was sent.
```

The question is if AgentDescription must be included in every StatusReport. We need to consider what happens both for WebSocket and plain HTTP transports.

tigrannajaryan:
Some thoughts on this topic:

- The Server typically keeps a state per each Agent. When the state changes on the Agent side it needs to be synced from the Agent to the Server.
- OpAMP allows the Agent to omit unchanged data in the AgentToServer message, since otherwise every AgentToServer can be potentially very large.
- For a variety of reasons we cannot expect that the Server always has the latest up-to-date state of each Agent (e.g. the Server may keep the state in RAM and a Server restart may loose the state).

WebSocket specifics:
1. WebSocket connections are persistent. We can assume that if the connection is open and maintained then all AgentToServer messages sent via that connection were processed by the Server and the cumulative state described by all those AgentToServer messages is correctly stored by the Server.
2. If the Server loses the state of the Agent it MUST NOT keep the WebSocket connection open, so that the assertion in point 1 above is valid.

Plain HTTP specifics:
1. The connections are not persistent (they may be kept alive, but that is unrelated to this discussion).
2. We cannot make assumptions about how long the Server keeps the Agent's state and whether the state is lost between AgentToServer messages sent by this Agent using HTTP requests.

Give the above, I believe the following is necessary:
- For plain HTTP transport every AgentToServer message must either include the full state of the Agent or it must include a hash of that state (the
state data may be broken down into smaller parts, and a hash of each part may be included). 
- We could have a different rule for WebSocket connections because of their persistent nature, however for simplicity of
[...]

PR #79, opened 2022-05-02
Clarify how state hashes are used between the Agent and the Server

tigrannajaryan:
Resolves https://github.com/open-telemetry/opamp-spec/issues/76

- Added a section that explains the state synchronization from
  the Agent to the Server.
- Added flags for Server to ask for AgentDescription and RemoteConfigStatus
  and dded hash field to AgentDescription and RemoteConfigStatus.
  This makes all 4 state messages consistently use the same state
  synchronization approach.

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -27,6 +27,7 @@ Note 2: this document requires a simplification pass to reduce the scope, size a
       - [agent_disconnect](#agent_disconnect)
       - [flags](#flags)
   * [ServerToAgent Message](#servertoagent-message)
+  * [Agent to Server State Synchronization](#agent-to-server-state-synchronization)
       - [instance_uid](#instance_uid-1)
       - [error_response](#error_response)
       - [remote_config](#remote_config)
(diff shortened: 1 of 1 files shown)


specification.md
```diff
@@ -483,6 +485,47 @@ message ServerToAgent {
...
+The compression is done by omitting all fields in these messages, except
+the hash field which MUST always be present (see below for how the hash field is used).
+If any of the fields in the message has changed then the compression cannot be used
+and all fields MUST be present.
+
+If all AgentToServer messages are reliably delivered to the Server and the Server
+correctly processes them then such compression is safe and the Server should always
+have the correct latest state of the Agent. 
+
+However, it is possible that the Agent and Server loose the synchronization and the Agent
```
yurishkuro, line 510: loose or lose?

tigrannajaryan: `lose`. 
Thanks for the catch, fixed.

specification.md
```diff
@@ -896,11 +931,28 @@ The AgentDescription message has the following structure:
...
+    repeated KeyValue non_identifying_attributes = 3;
 }
 ```
 
+#### hash
+
+The hash of the content of all other fields (even if the other fields are omitted
+for compression).
+
+If the content of the other fields did not change since it was last reported the Agent is
```
pmm-sumo, line 945: Maybe the general hash explanation could be put in a "Agent To Server Synchronization" subsection and the details on it could be referenced here and in other places where hash is present (rather than refer to Agent Description hash)? I think that would make for a more clear structure

tigrannajaryan: Re-worded. PTAL.

pmm-sumo (approved):

specification.md
```diff
@@ -483,6 +485,54 @@ message ServerToAgent {
 }
 ```
 
+## Agent to Server State Synchronization
```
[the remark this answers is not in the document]
andykellr: I think the text is clear.

andykellr (approved):

specification.md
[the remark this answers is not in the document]
andykellr: I think the text is clear.

PR #81, opened 2022-05-13
Consistently capitalize "Agent" and "Server"

tigrannajaryan:


specification.md
```diff
@@ -320,7 +320,7 @@ to poll the Server like it is done when using the HTTP transport (see below).
 ## Plain HTTP Transport
 
 The second supported transport for OpAMP protocol is plain HTTP connection. The
-Agent is an HTTP client and the Server is an HTTP client server. The Agent makes
+Agent is an HTTP client and the Server is an HTTP server. The Agent makes
```
pmm-sumo, line 323: Not sure, but perhaps for consistency:
```suggestion
Agent is an HTTP client and the Server is an HTTP Server. The Agent makes
```

specification.md
```diff
@@ -761,7 +761,7 @@ If the status report is processed successfully by the Server then the
 be populated as necessary.
 
 Here is the sequence diagram that shows how status reporting works (assuming
-server-side processing is successful):
+Server-side processing is successful):
```
pmm-sumo, line 764: In this instance, I think it's weird to have capitalized "Server-side"

```suggestion
server-side processing is successful):
```

specification.md
tigrannajaryan: I intentionally skipped this. I use capital "S" to denote the OpAMP Server. The phrase "HTTP server" refers to any server.

pmm-sumo (approved):
I think that the capitalisation depends if it's a proper noun or not. I.e. when we refer to "OpAMP Server" or "OpAMP Agent" it's ok to have it capitalized, but when used in a very broad meaning (any server, any agent) it's not. I reviewed and added some comments when I think it should be not capitalized

Having said that, I'm not a native speaker and perhaps it should be done differently

specification.md
tigrannajaryan: I agree. Reverted.

tigrannajaryan:
> I think that the capitalisation depends if it's a proper noun or not. I.e. when we refer to "OpAMP Server" or "OpAMP Agent" it's ok to have it capitalized, but when used in a very broad meaning (any server, any agent) it's not.

Yes, I completely agree.

specification.md
```diff
@@ -2664,7 +2664,7 @@ WebSocket is a bidirectional, message-oriented protocol that uses plain HTTP for
 establishing the connection and then uses the HTTP's existing TCP connection to
 deliver messages. It has been an
 [RFC](https://datatracker.ietf.org/doc/html/rfc6455) standard for a decade now.
-It is widely supported by browsers, servers, proxies and load balancers, has
+It is widely supported by browsers, Servers, proxies and load balancers, has
```
pmm-sumo, line 2667: I think it's not a proper noun here:
```suggestion
It is widely supported by browsers, servers, proxies and load balancers, has
```

tigrannajaryan: Reverted.

Issue #84, opened 2022-05-20
Move messages in StatusReport to AgentToServer message


tigrannajaryan:
It is not clear why we have StatusReport, which contains 4 fields, instead of putting the fields directly in AgentToServer.

Can we move the fields to AgentToServer and eliminate StatusReport message?

tigrannajaryan:
Here is what the change can look like: https://github.com/open-telemetry/opamp-go/pull/83

@andykellr @pmm-sumo what do you think?

andykellr:
I like this change.

PR #85, opened 2022-05-25
Move messages in StatusReport to AgentToServer message

tigrannajaryan:
- StatusReport was an unnecessary message, remaining from the time
  when we had DataToAgent which had a oneof field, and it was
  impossible to set multiple fields that together form the status.
  All its fields are moved to AgentToServer message now and can be
  set directly there (there is no longer a oneof limitation so now
  this is possible to do).
- Moved Agent to Server State Synchronization under Status Reporting
  where it is more relevant and renamed to Agent Status Compression.

Resolves https://github.com/open-telemetry/opamp-spec/issues/84


andykellr (approved):
I like this change.

PR #89, opened 2022-06-15
Enclose the TOC in a details element.

jpeach:
It's helpful to have a TOC, but there's a fair amount of scrolling to
do before you get to the text. If we enclose it in a details element,
the reader can show the TOC only when then need it.

Signed-off-by: James Peach <jpeach@cloudflare.com>

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -14,6 +14,9 @@ Note 2: this document requires a simplification pass to reduce the scope, size a
 
 #
 
+<details>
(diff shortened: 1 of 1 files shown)


tigrannajaryan (approved):

tigrannajaryan:
@jpeach Thanks, this is a good change but markdown-toc-check action is failing. Maybe it needs some other settings, see if you can fix it.

Issue #101, opened 2022-06-29
Status compression is complicated and confusing


tigrannajaryan:
[Status compression](https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#agent-status-compression) is quite complicated and confusing. It uses 4 different hashes and lots of code to enable compression and detection of lost messages (during server restarts).

We can simplify this significantly by doing this instead:
- Eliminate all "hash" fields.
- Allow omitting the sub-messages of AgentToServer message. When omitted it is implied that previously reported value of the sub-message is current (unchanged).
- To detect lost messages have one auto-incremented sequence_num field AgentToServer message. Server can easily detect losses by just keeping the last sequence_num (as opposed to keeping 4 different hashes).

Here is a draft PR demonstrate how it works: https://github.com/open-telemetry/opamp-go/pull/93
It is more than 100 lines of code removed.

tigrannajaryan:
@andykellr @pmm-sumo I am doing a simplification pass over the spec in order to remove unnecessary complexity.

The whole hashes and status compression is unnecessarily complicated. I made a PR with a simpler approach.

Please let me know what you think.

PR #102, opened 2022-06-29
Simplify status compression

tigrannajaryan:
Resolves https://github.com/open-telemetry/opamp-spec/issues/101

Status compression was quite complicated and confusing. It used 4 different
hashes and lots of code to enable compression and detection of lost messages
(during server restarts).

We simplified this significantly by doing this instead:
- Eliminate all "hash" fields.
- Allow omitting the sub-messages of AgentToServer message. When omitted it is implied that previously reported value of the sub-message is current (unchanged).
- To detect lost messages have one auto-incremented sequence_num field AgentToServer message. Server can easily detect losses by just keeping the last sequence_num (as opposed to keeping 4 different hashes).

Here is a PR that demonstrates how it works:
https://github.com/open-telemetry/opamp-go/pull/93

It is more than 100 lines of code removed.


specification.md
```diff
@@ -878,43 +875,30 @@ believes the Server has the latest data while in reality the Server doesn't. Thi
...
+In order to detect this situation and recover from it, AgentToServer message
+contains a sequence_num field. The field is an integer number that is incremented
+every time the Agent has a new AgentToServer message to send.
 
-The Server SHOULD store the received hash value for each message type. When the Server
-receives a message of the same type with a hash value that is different from the last
-stored hash and with omitted data then the Server knows it does not have full status
-of this particular message.
+When the Server receives an AgentToServer message sequence_num field value that is not
+exactly by one greater than the previously received sequence_num value then the Servemaker
```
andykellr, line 883:
```suggestion
exactly by one greater than the previously received sequence_num value then the Server
```

specification.md
```diff
@@ -878,43 +875,30 @@ believes the Server has the latest data while in reality the Server doesn't. Thi
 possible for example if the Server is restarted while the Agent keeps running and sends
 AgentToServer messages, which the Server does not receive because it is temporarily down.
 
-In order to detect this situation and recover from it, every compressible message
-contains a hash field. The field is the hash of the content of every other field.
-The hash is computed on full, uncompressed message (as if no compression is used) and
-then unchanged fields may be omitted from the message. Note that either all fields in the
-message must be present or all fields (except hash) must be omitted.
+In order to detect this situation and recover from it, AgentToServer message
```
andykellr, line 878:
```suggestion
In order to detect this situation and recover from it, the AgentToServer message
```

specification.md

specification.md

andykellr (approved):

PR #105, opened 2022-07-07
Remove unnecessary attributions

tigrannajaryan:
Github already shows contributors. At Otel we don't typically include
authors and contributors in the document text itself.


andykellr (approved):
Makes sense. Thank you for all of your hard work on the protocol!

Issue #113, opened 2022-07-26
Require that instance_uid is ULID instead of only recommending it


tigrannajaryan:
The [spec says](https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#agenttoserverinstance_uid):

>The instance_uid field is a globally unique identifier of the running instance of the Agent. The Agent SHOULD self-generate this identifier and make the best effort to avoid creating an identifier that may conflict with identifiers created by other Agents. The instance_uid SHOULD remain unchanged for the lifetime of the Agent process. The recommended format for the instance_uid is [ULID](https://github.com/ulid/spec).

The "recommended" phrase is equivalent to `SHOULD`, however weaker than `MUST`. I think we need to change this to be `MUST`:
- Avoid mistakes when the Agent chooses too short instance id and causes collisions.
- Make the Server's life a bit easier since will know it is always a 128 bits of data and can store in places where UUIDs can be stored (e.g. Postgres has a data type UUID).

I don't see what do we gain by keeping it as only recommended.

tigrannajaryan:
@andykellr what do you think?

PR #116, opened 2022-07-28
Require that instance_uid is ULID instead of only recommending it

tigrannajaryan:
Resolves https://github.com/open-telemetry/opamp-spec/issues/113

This PR changes ULID to be a `MUST` in order to:
- Avoid mistakes when the Agent chooses too short instance id and causes collisions.
- Make the Server's life a bit easier since will know it is always a 128 bits of data and can store in places where UUIDs can be stored (e.g. Postgres has a data type UUID).


andykellr (approved):
This will be a breaking change for bindplane-op agents, but I think it is a good change.

tigrannajaryan:
> This will be a breaking change for bindplane-op agents, but I think it is a good change.

Yes. Good news is that this seems to be the last remaining known issue that will be a breaking change. We should probably wait for a bit after merging this one to see if any other issues are discovered that will be also breaking so that we can release them at once (multiple breaking releases are usually more painful - that's why we promise no more frequently than once every 3 months when in Beta).

PR #122, opened 2022-08-23
Separating the notions of Agent and Client

PeterF778:
Separating the previous notion of Agent into Agent (as a managed entity) and Client (as an endpoint of OpAMP).

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -183,6 +183,7 @@ mixed Agents from different vendors.
 
 OpAMP supports the following functionality:
 
+* Maintaining Agent identity.
 * Remote configuration of the Agents.
 * Status reporting. The protocol allows the Agent to report the properties of
   the Agent such as its type and version or the operating system type and
@@ -203,8 +204,12 @@ of a large fleet of mixed Agents (e.g. OpenTelemetry Collector, Fluentd, etc).
 
 # Communication Model
 
-The OpAMP Server manages Agents that implement the client-side of OpAMP
-protocol. The Agents can optionally send their own telemetry to an OTLP
+The OpAMP Server manages Agents that provide a client-side implementation of OpAMP
+protocol, further referred to as OpAMP Client or simply Client.
+OpAMP does not assume any particular relationship between the Agent and the Client,
+the Client can be run in a separate process, a sidecar, a plugin, or be fully integrated into the Agent code.
+
+The Agents can optionally send their own telemetry to an OTLP
 destination when directed so by the OpAMP Server. The Agents likely also connect
 to other destinations, where they send the data they collect:
 
@@ -245,27 +250,27 @@ instance_uid for short). The instance_uid is recorded in each message sent from
 the Agent to the Server and from the Server to the Agent.
 
 The default URL path for the connection is /v1/opamp. The URL path MAY be
-configurable on the Agent and on the Server.
+configurable on the Client and on the Server.
 
 ## WebSocket Transport
 
 One of the supported transports for OpAMP protocol is
-[WebSocket](https://datatracker.ietf.org/doc/html/rfc6455). The Agent is a
-WebSocket client and the Server is a WebSocket Server. The Agent and the Server
+[WebSocket](https://datatracker.ietf.org/doc/html/rfc6455). The OpAMP Client
+is a WebSocket client and the Server is a WebSocket Server. The Client and the Server
 communicate using binary data WebSocket messages. The payload of each WebSocket
 message is a
 [binary serialized Protobuf](https://developers.google.com/protocol-buffers/docs/encoding)
-message. The Agent sends AgentToServer Protobuf messages and the Server sends
+message. The Client sends AgentToServer Protobuf messages and the Server sends
 ServerToAgent Protobuf messages:
 
 ```
-        ┌───────────────┐                        ┌──────────────┐
-        │               │      AgentToServer     │              │
-        │               ├───────────────────────►│              │
-        │     Agent     │                        │    Server    │
-        │               │      ServerToAgent     │              │
-        │               │◄───────────────────────┤              │
-        └───────────────┘                        └──────────────┘
+        ┌────────────\ \────────┐                        ┌──────────────┐
+        │            / /        │      AgentToServer     │              │
+        │            \ \ OpAmp  ├───────────────────────►│              │
+        │     Agent  / /        │                        │    Server    │
+        │            \ \ Client │      ServerToAgent     │              │
+        │            / /        │◄───────────────────────┤              │
+        └────────────\ \────────┘                        └──────────────┘
 ```
 
 Typically a single Server accepts WebSocket connections from many Agents. Agents
@@ -281,11 +286,11 @@ sequence of messages exchanged by the Agent and the Server is defined for each
 particular capability in the corresponding section of this specification.
 
 The sequence is normally started by an initiating message triggered by some
-external event. For example after the connection is established the Agent sends
+external event. For example after the connection is established the Client sends
 a AgentToServer message. In this case the "connection established" is the
 triggering event and the AgentToServer is the initiating message.
 
-Both the Agent and the Server may begin a sequence by sending an initiating
+Both the Client and the Server may begin a sequence by sending an initiating
 message.
 
 The initiating message may trigger the recipient to send one or more messages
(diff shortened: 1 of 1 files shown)


specification.md
```diff
@@ -183,6 +183,7 @@ mixed Agents from different vendors.
 
 OpAMP supports the following functionality:
 
+* Maintaining Agent identity.
```
tigrannajaryan, line 186: I think this is already covered by the "Status reporting" below. It says "allows the Agent to report the properties of
the Agent".

specification.md
```diff
@@ -245,27 +250,27 @@ instance_uid for short). The instance_uid is recorded in each message sent from
...
 One of the supported transports for OpAMP protocol is
-[WebSocket](https://datatracker.ietf.org/doc/html/rfc6455). The Agent is a
-WebSocket client and the Server is a WebSocket Server. The Agent and the Server
+[WebSocket](https://datatracker.ietf.org/doc/html/rfc6455). The OpAMP Client
+is a WebSocket client and the Server is a WebSocket Server. The Client and the Server
 communicate using binary data WebSocket messages. The payload of each WebSocket
 message is a
 [binary serialized Protobuf](https://developers.google.com/protocol-buffers/docs/encoding)
-message. The Agent sends AgentToServer Protobuf messages and the Server sends
+message. The Client sends AgentToServer Protobuf messages and the Server sends
```
tigrannajaryan, line 263: Shouldn't we also rename AgentToServer to ClientToServer?

specification.md
```diff
@@ -1756,18 +1761,18 @@ Config    Config │ ServerToAgent{AgentRemoteConfig}  │   │ Fetch   │
...
 
 The Server may also initiate sending of a remote configuration on its own,
-without waiting for a status report from the Agent. This can be used to
+without waiting for a status report from the Client. This can be used to
 re-configure an Agent that is connected but which has nothing new to report. The
 sequence diagram in this case looks like this:
 
 ```
-               Agent                              Server
+    Agent      Client                             Server
```
tigrannajaryan, line 1775: I understand that this aims to show that a particular action (Config Merging) is done by the Agent and not the Client. I think it is a good idea to show that, but the lack of the vertical line (like it is usually done for sequence diagrams) makes it a bit confusing.

tigrannajaryan (commented):
Thanks @PeterF778 
This is very useful. I am not entirely sure the renamings work fine in all places. I wonder also what @andykellr thinks.

Also, no changes in the Packages section? Do you think it should stay as is?

specification.md
PeterF778: Perhaps my wording is not right, but I wanted to say that the protocol has provisions for maintaining unique identities of the Agent instances. I think it is a small but very important feature and it goes beyond reporting properties.

specification.md
PeterF778: No, the protocol is still very much Agent-oriented. The Server is almost exclusively concerned about the Agent capabilities, configuration and status, and hardly sees the Client as an independent entity - even when asking for Agent upgrade, it happens magically behind the scene, even though we know that it requires some sort of a supervisor.

specification.md
PeterF778: Yes, this is a bit confusing. Two vertical lines, one for Agent and one for Client might be misleading as well, by strongly suggesting that these have to be separate entities.
Perhaps we could use a solid vertical line for the Agent, and a dashed vertical line for the Client. I do not care too much that it is a non-standard way of showing things, if it proves to be helpful.

PeterF778:
Thanks @tigrannajaryan for reviewing.
I tried to be careful with the changes, but it is a long document, and I might have missed some places. Yes, I looked carefully at the Packages section, and I thought all was good there.
I'll have a small update soon.

portertech:
Thank you for this :+1: I recently found myself stumbling over words when explaining how OpAMP works with/within the OT Agent :sweat_smile:

tigrannajaryan:
> Thank you for this 👍 I recently found myself stumbling over words when explaining how OpAMP works with/within the OT Agent 😅

@portertech do you think the changes introduced by this PR make it clearer?

specification.md
```diff
@@ -1655,22 +1656,22 @@ The Agent SHOULD periodically report its metrics to the destination offered in t
...
-                  │                                      │  Backend
-                  │ServerToAgent{ConnectionSettingsOffer}│
-                  │◄─────────────────────────────────────┤    │
-                  │                                      │    │
-                  │                                           │
+        │         │                                      │  Backend
+        │         │ServerToAgent{ConnectionSettingsOffer}│
+        ┌─────────│◄─────────────────────────────────────┤    │
+        │         │                                      │    │
+        ▼         │                                           │
```
tigrannajaryan, line 1665: I think we can delete the vertical line for Client on lines 1665-1676. This will eliminate the intersection with horizontal arrows on lines 1668 and 1674.

specification.md
```diff
@@ -1719,28 +1720,28 @@ Configuration with other inputs available to the Agent, e.g. a locally available
...
-                 │                                   │   │ Status  │
+     │           │ AgentToServer{}                   │   ┌─────────┐
+     │           ├──────────────────────────────────►├──►│ Process │
+     │           │                                   │   │ Status  │
 Local     Remote │                                   │   │ and     │
 Config    Config │ ServerToAgent{AgentRemoteConfig}  │   │ Fetch   │
-  │     ┌────────┤◄──────────────────────────────────┤◄──┤ Config  │
-  ▼     ▼        │                                   │   └─────────┘
+  │  │  ┌────────┤◄──────────────────────────────────┤◄──┤ Config  │
+  ▼  │  ▼        │                                   │   └─────────┘
```
tigrannajaryan, line 1744: Can we make the vertical line for the Agent continuous? Probably "Local Config" to the left a bit to make room for the vertical line.

tigrannajaryan:
@andykellr @portertech I wonder what you guys think about this change. As the author of the current text I am probably biased towards the usage of "Agent" everywhere, so I may not be the best person to tell if this change makes the spec more clear (I think it definitely does in some places, but I am worried that overall it may result in some confusion because of using 2 different concepts).
It would be great to know your opinion.

specification.md
andykellr: I think the pattern in the `ServerToAgent{ConnectionSettingsOffer}` is fine but we should make a little extra room to allow the vertical line to continue on lines 1776/1777. I don't think this implies that they must be separate entities. This is discussed in the `Communication Model` section and I think it is clear.

andykellr (commented):
I agree with the issues Tigran highlighted and my only other piece of feedback was a small grammatical change.

Thank you for taking the time to go through the spec and identify all of the necessary changes. I expanded all of the unchanged sections as I review the changes and I believe all of the Agent => Client changes are appropriate and I did not find any that were overlooked.

PeterF778:
Thank you @andykellr for the review. I agree with yours and @tigrannajaryan suggestions. I'll make the corrections next week.

specification.md
PeterF778: Yes, I agree.

specification.md
```diff
@@ -301,60 +300,60 @@ separation between "request" and "response" messages types. The role of the
...
+Client initiates an HTTP request when it has an AgentToServer message to deliver.
 The Server responds to each HTTP request with a ServerToAgent message it wants
 to deliver to the Agent. If the Agent has nothing to deliver to the Server the
-Agent MUST periodically poll the Server by sending an AgentToServer message
+Client MUST periodically poll the Server by sending an AgentToServer message
 where only [instance_uid](#agenttoserverinstance_uid) field is set. This gives the Server an
 opportunity to send back in the response any messages that the Server wants to
 deliver to the Agent (such as for example a new remote configuration).
 
 The default polling interval when the Agent does have anything to deliver is 30
```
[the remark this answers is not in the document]
PeterF778: Yes, good catch.

specification.md
PeterF778: Yes, it will look better!

tigrannajaryan (approved):

tigrannajaryan:
@andykellr I will keep this open for you to also review.

andykellr (approved):

tigrannajaryan:
Merged. Thanks @PeterF778 !

PR #126, opened 2022-09-22
Move proto files to opamp-spec repo

tigrannajaryan:
Contributes to https://github.com/open-telemetry/opamp-go/issues/133

Now that we are adding OpAMP Java implementation the protos can no longer live in opamp-go. We are moving the protos to the spec repo since it fits this repo's purpose naturally.

The proto files are copied verbatim from opamp-go. Any changes needed in the proto files can be done in a future PR.

In the future it is also worth exploring if we can delete the message definitions from specification.md and instead link them directly to the proto files so that we have a single source of truth.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,14 @@
+# Function to execute a command.
+# Accepts command to execute as first parameter.
+define exec-command
+$(1)
+
+endef
+
 ALL_DOCS := $(shell find . -type f -name '*.md' -not -path './.github/*' -not -path './node_modules/*' | sort)
 
 .PHONY: all
-all: markdown-toc markdown-link-check markdownlint
+all: markdown-toc markdown-link-check markdownlint gen-proto
 
 
 # This target runs markdown-toc on all files that contain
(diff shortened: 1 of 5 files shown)


tigrannajaryan:
@jlegoff please review.

Makefile
```diff
@@ -40,3 +47,27 @@ markdownlint:
...
+PROTO_FILES := $(wildcard proto/*.proto)
+
+PROTO_GEN_GO_DIR ?= $(GENDIR)/go
+
+gen-proto: gen-go
+.PHONY: gen-proto
+
+# Generate Protobuf Go files.
+.PHONY: gen-go
+gen-go:
```
jlegoff, line 63: I tried running this command, but the docker image doesn't exist for my platform (linux/arm64/v8)

Is the purpose of this rule to check that the go files are correctly generated?

andykellr (approved):
I'm unable to confirm that `make gen-proto` works because it does not support M1 (see https://github.com/open-telemetry/build-tools/issues/42). Otherwise, this looks good.

jlegoff (approved):
I also cannot check the gen-go rule, but otherwise this looks good to me 👍

Makefile
tigrannajaryan: Yes, that's the purpose. It is using the same docker image as https://github.com/open-telemetry/opentelemetry-proto

tigrannajaryan:
I forgot to add the github. Done now. The action runs and checks that the protos can be generated in Go.

Issue #49, opened 2021-12-07
Add ability to send custom messages


tigrannajaryan:
It would be nice if there was a way to allow he Agent and Server to exchange arbitrary
messages over the same OpAMP communication channel in a safe manner without disrupting
the protocol's functionality.

This would allow extending OpAMP applications to cases where Agent and Server need to do
something extra that is not supported by OpAMP but which Agent and Server can agree upon
upfront and use OpAMP channel instead of having yet another communication chanell
which implements all the transport machinery again.

One possible simplistic way would be to include an extra `bytes`-type field in AgentToServer and ServerToAgent
which the Agent and Server can agree to use however they want (e.g. to encode a custom Protobuf message
in the `bytes`).

tigrannajaryan:
@andykellr FYI.

tigrannajaryan:
Resolved by https://github.com/open-telemetry/opamp-spec/pull/132

PR #132, opened 2022-09-27
Add CustomMessage capability

andykellr:
See https://github.com/open-telemetry/opamp-spec/issues/49

If we agree on the direction of this, I plan to do the following:

* updates to this opamp-spec PR:
  * add protos to the opamp.proto file
* updates to the opamp-go repo in a separate PR:
  * add OnCustomMessage method to the client Callbacks interfaces
  * add SendCustomMessage to the OpAMPClient
  * add implementation to the client and server
  * add tests for that implementation


andykellr:
@tigrannajaryan please review and I will update and fill in the rest. Thanks!

andykellr:
Related to this, should we enforce `line-length` in .markdownlint.yaml? My editor wraps at 120. I noticed most of the existing lines are between 80-90.

specification.md
```diff
@@ -2201,6 +2219,169 @@ authenticity of the downloaded file, for example can be the
...
+
+While interoperability between many Agents and Agent Management Servers is a goal of OpAMP, additional custom
+functionality may be only supported by specific Agents and Servers. Where appropriate, implementers are encouraged to
+document their use of CustomMessage so that support can be added to additional platforms.
+
+### CustomMessage
+
+CustomMessage allows for custom messages to be sent between the Agent and Server. It is supported on ServerToAgent and
+AgentToServer. It requires that the Agent and Server both agree on the contents and encoding of the message. If the type
+is not recognized, the message MUST be ignored and a reply with CustomMessageAck with Status NOT_SUPPORTED MUST be sent.
```
cpheps, line 2241: For some reason the `MUST be ignored` part of this line seems a bit heavy handed to me. I'm wondering if it's fine to just say `MUST replay with CustomMessageAck with Status NOT_SUPPORTED`. Not sure if we need to explicitly call out the receiver must ignore the message.

andykellr: My intent was to make it clear that it is not an error, but I could modify the sentence.

cpheps: Since there are only three options for a Status I think the fact that it's called out it has to be `NOT_SUPPORTED` rather than `ERROR` conveys this.

specification.md
```diff
@@ -2201,6 +2219,169 @@ authenticity of the downloaded file, for example can be the
...
+
+This Ack is in response to the CustomMessage with the corresponding sequence_num.
+
+#### CustomMessageAck.status
+
+The status field indicates if and how the custom message was handled. Possible values are:
+
+NOT_SUPPORTED: The CustomMessage was ignored because the message type was not recognized or supported. The error_message may contain additional information.
+
+OK: The CustomMessage was handled successfully.
```
cpheps, line 2305: Should we expect the `error_message` to be empty always in this case?

specification.md
```diff
@@ -2201,6 +2219,169 @@ authenticity of the downloaded file, for example can be the
...
+
+ERROR: The CustomMessage was unable to be handled successfully. The error_message should contain additional information.
+
+#### CustomMessageAck.error_message
+
+Error message in string form, typically human readable.
```
cpheps, line 2311: Seems like this format is arbitrary and left up to agreement based on the specific Agent/Server? Only asking because we call out `typically human readable` which might not be necessary if the receiver of the Ack knows how to parse it and can present it to the user in a useful way.

tigrannajaryan:
> Related to this, should we enforce `line-length` in .markdownlint.yaml? My editor wraps at 120. I noticed most of the existing lines are between 80-90.

Yes, please keep the line wraps similar to the rest of the spec (I think it is at 90).

andykellr:
I made significant changes to this proposal and have some time to work on the implementation. @tigrannajaryan please have a look.

specification.md
```diff
@@ -2474,6 +2530,134 @@ authenticity of the downloaded file, for example can be the
...
+```protobuf
+message CustomMessage {
+    string type = 1;
+    bytes data = 2;
+}
+```
+
+##### CustomMessage.type
+
+The message type should a reverse FQDN that uniquely identifies the custom message type.
```
jaronoff97, line 2589:
```suggestion
The message type should have a reverse FQDN that uniquely identifies the custom message type.
```

specification.md
```diff
@@ -2474,6 +2530,134 @@ authenticity of the downloaded file, for example can be the
...
+for which the Agent could collect telemetry. In this example the Server will send a
+message to the Agent to ask for available services. It will expect a response from the
+Agent containing information about the available services.
+
+###### Agent Connection
+
+On connection, the Agent sends a CustomMessageCapabilities message including the message
+type "io.opentelemetry.find_services". In response, the Server sends a
+CustomMessageCapabilities message including the message type
+"io.opentelemetry.find_services_response".
```
jaronoff97, line 2645: In this scenario, the server specifies `io.opentelemetry.find_services_response` which isn't one of the capabilities listed by the agent. Wouldn't we expect the agent to broadcast that its custom capabilities include both `io.opentelemetry.find_services_response` and `io.opentelemetry.find_services`?

jaronoff97 (approved):
Had one question, overall I'm a fan of this change. It would be great for the future to support custom messages to allow the protocol to be extended upon more easily. For the operator group, I could imagine the bridge allowing for a message for restarting a workload or even pausing telemetry as in the examples.

specification.md
andykellr: That's a good question and maybe a little confusing. The agent supports the request and the server supports the response. Maybe the custom capability should be separate and not based on the message, e.g. `io.opentelemetry.discovery` is supported by both client and server and the corresponding message types are implied.

andykellr:
> > Related to this, should we enforce `line-length` in .markdownlint.yaml? My editor wraps at 120. I noticed most of the existing lines are between 80-90.
> 
> Yes, please keep the line wraps similar to the rest of the spec (I think it is at 90).

I attempted to enforce this with the following change, but it requires many sections to be rewrapped to meet the requirement. I can do this in a separate PR if we believe it should be enforced.

```
<!-- markdownlint-configure-file
{
  "line-length": {
    "line_length": 90,
    "stern": true
  }
}
-->
```

specification.md
```diff
@@ -618,6 +640,22 @@ the Agent initiates the creation of connection settings.
 
 See [ConnectionSettingsRequest](#connectionsettingsrequest-message) message for details.
 
+##### AgentToServer.custom_message_capabilities
+
+Status: [Development]
+
+A message indicating custom message types supported by Agent.
```
evan-bradley, line 647:
```suggestion
A message indicating custom message types supported by an Agent.
```

specification.md
```diff
@@ -2474,6 +2530,134 @@ authenticity of the downloaded file, for example can be the
...
+supported on ServerToAgent and AgentToServer to allow for custom communication in both
+directions. It requires that the Agent and Server both agree on the contents and encoding
+of the messages. If the message type is not recognized, the message can be ignored.
+
+A CustomMessage identifies the type of message and contains binary data with the contents
+of the message. The format of the data will depend on the message type and is outside of
+the scope of the OpAMP protocol.
+
+```protobuf
+message CustomMessage {
```
evan-bradley, line 2581: If we think it would be helpful, how do we suggest that custom message authors implement message versioning? For a pause message, should the version be in the name with something like `io.opentelemetry.pause/v2`, or should versioning generally happen in the data field? Alternatively, should there be `version` field on `CustomMessage`? Ultimately anyone writing custom messages is free to use their preferred versioning strategy, but I think some guidelines could help encourage consistency and interoperability.

specification.md
evan-bradley: I have to admit that I also had to take a minute to figure this out at first. I like that specifying supported custom messages mirrors the way capabilities work in OpAMP and that it makes it very clear what the Agent and Server support. If we separate the capability from the message, I think it would make sense in `CustomMessageCapabilities` to indicate the capability (`io.opentelemetry.discovery`) then a list of messages the peer supports receiving as part of that capability, (`request` for the Agent and `response` for the Server) and possibly the messages it may send (`response` for the Agent and `request` for the Server).

specification.md
```diff
@@ -2474,6 +2530,134 @@ authenticity of the downloaded file, for example can be the
...
+protocol.
+
+While interoperability between many Agents and Agent Management Servers is a goal of
+OpAMP, additional custom functionality may only be supported by specific Agents and
+Servers. Where appropriate, implementers are encouraged to document their use of
+CustomMessage so that support can be added to additional platforms.
+
+#### CustomMessageCapabilities Message
+
+The Agent and Server should both use this message to signal that they support specific
```
portertech, line 2553: Just a thought. Although probably unlikely, a string capability for one server may differ from the functionality on another (e.g. vendor) and a connecting agent may not know any better resulting in unexpected behaviour. Do we care about interoperability? Do the strings need to be distro/vendor specific? Do we need another field?

andykellr: The intent is that they would use a reverse FQDN to be vendor specific, e.g. our pause/resume feature might be `com.observiq.pause`. I used `io.opentelemetry` in the examples to keep vendor specific names out of the spec.

specification.md
andykellr: Because the version may affect the decoding of the data field, any version information should be in the message type. I think adding `/version` to the name makes sense.

I'm going to change things slightly to add a `capability` field to CustomMessage and will document this in the name.

specification.md
evan-bradley: I like `CustomCapabilities` if we're going to say that supporting a capability implies full message support. The name lines up well with existing OpAMP terminology and is a bit more to the point.

In regards to specifying supported messages or suggesting that peers support all messages for a capability, I think that it's okay to just imply full support given that we are versioning the capabilities and considering a custom capability will be used for a very specific flow. It would also make things easier to reason about during implementation.

andykellr:
I just updated the PR with `CustomCapabilities` in addition to `CustomMessage` as discussed above.

proto/opamp.proto
```diff
@@ -817,3 +833,28 @@ message AgentConfigFile {
...
+}
+
+message CustomMessage {
+    // A reverse FQDN that uniquely identifies the custom capability.
+    // Status: [Development]
+    string capability = 1;
+
+    // The custom message type.
+    // Status: [Development]
+    string type = 2;
```
djaglowski, line 855: I may be missing context on this but would it make sense to just combine capability and type into one field? It seems like `capability.type` is just as useful and a little simpler.

andykellr: I wanted the capability to match the one in the CustomCapabilities message.

djaglowski: I see, a particular capability is associated with a set of message types. Those types are _implied_ rather than declared.

specification.md
andykellr: I had something like that in the original proposal but I prefer to leave message IDs out of the custom message spec. If a capability requires additional request context in a response, it is free to provide that in the data payload of the response.

evan-bradley: If we had sequence numbers in both the `AgentToServer` and `ServerToAgent` messages, the OpAMP client/server libraries could automatically handle this to some degree for custom messages that are synchronously processed. Asynchronous handling of custom messages would still require it to be put into the message data.

andykellr: I think we should add it when we have a good use case that requires it and start with an implementation within the message and then see if we should factor that out to be used by other implementations.

evan-bradley: Agreed, I think that is separate from this change.

specification.md
```diff
@@ -2474,6 +2531,194 @@ authenticity of the downloaded file, for example can be the
...
+###### Agent Connection
+
+On connection, the Agent sends an AgentToServer message with CustomCapabilities including
+the capabilities "io.opentelemetry.pause" and "io.opentelemetry.resume". In response, the
+Server sends a ServerToAgent message with a CustomCapabilities message including
+"io.opentelemetry.pause" and "io.opentelemetry.resume".
+
+```
+CustomCapabilities {
+  capabilities: ["io.opentelemetry.pause", "io.opentelemetry.resume"]
```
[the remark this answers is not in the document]
andykellr: I updated the example to use `pause` and `resume` messages in ServerToAgent and `status` and `error` messages in AgentToServer. I think `error` messages will be common and likely share the same structure, but I'm hesitant to add more constraints to the design at this point.

evan-bradley (approved):
Overall this looks good to me, I only have minor comments on wording.

specification.md
```diff
@@ -2474,6 +2531,205 @@ authenticity of the downloaded file, for example can be the
...
+```
+
+##### CustomMessage.capability
+
+A reverse FQDN that uniquely identifies the capability and matches one of the capabilities
+in the CustomCapabilities message.
+
+##### CustomMessage.type
+
+The custom message type.
```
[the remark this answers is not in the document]
andykellr: Good catch. I meant to come back to this one. I updated most of the field descriptions.

proto/opamp.proto
```diff
@@ -817,3 +833,33 @@ message AgentConfigFile {
...
+    // A reverse FQDN that uniquely identifies the capability and matches one of the
+    // capabilities in the CustomCapabilities message.
+    // Status: [Development]
+    string capability = 1;
+
+    // Type of message within the capability. The capability defines the types of custom
+    // messages that are used to implement the capability. The type must only be unique
+    // within the capability.
+    // Status: [Development]
+    string type = 2;
```
tigrannajaryan, line 859: Is the purpose of `type` to aid in decoding `data` field? Otherwise, we could embed the `type` information into the `data`.

proto/opamp.proto
```diff
@@ -817,3 +833,33 @@ message AgentConfigFile {
...
+    // and should match a capability specified in a supported CustomMessage.
+    // Status: [Development]
+    repeated string capabilities = 1;
+}
+
+message CustomMessage {
+    // A reverse FQDN that uniquely identifies the capability and matches one of the
+    // capabilities in the CustomCapabilities message.
+    // Status: [Development]
+    string capability = 1;
```
tigrannajaryan, line 853: So, if an interaction involves multiple capabilities it requires multiple AgentToServer/ServerToAgent messages, right?.

proto/opamp.proto
```diff
@@ -180,6 +188,14 @@ message ServerToAgent {
     // will be performed.
     // Status: [Beta]
     ServerToAgentCommand command = 9;
+
+    // A message indicating custom capabilities supported by the Server.
+    // Status: [Development]
+    CustomCapabilities custom_capabilities = 10;
```
tigrannajaryan, line 194: Is it possible to omit this? With a long list of capabilities this can become significant payload size-wise.

specification.md
```diff
@@ -2474,6 +2531,207 @@ authenticity of the downloaded file, for example can be the
 The exact signing and verification method is Agent specific. See
 [Code Signing](#code-signing) for recommendations.
 
+### Custom Messages
```
tigrannajaryan, line 2534:
```suggestion
### Custom Messages

Status: [Development]
```

specification.md
```diff
@@ -2474,6 +2531,207 @@ authenticity of the downloaded file, for example can be the
...
+
+#### Motivation
+
+The OpAMP protocol is intended to cover all aspects of remote management and configuration
+of Agents. If more requirements within this core functionality are identified, it is
+expected that the protocol will be extended to support this additional behavior in future
+releases. However, there are some use cases that require communication with Agents that
+will not by supported by the OpAMP protocol because they are not within the scope of Agent
+remote management. CustomCapabilities and CustomMessage allow custom behavior to be
```
tigrannajaryan, line 2543:
```suggestion
remote management or are highly specific to one agent and not generalizable to other agent types. CustomCapabilities and CustomMessage allow custom behavior to be
```

specification.md
```diff
@@ -2474,6 +2531,207 @@ authenticity of the downloaded file, for example can be the
...
+
+#### CustomCapabilities
+
+The Agent and Server should both use this message to signal that they support specific
+custom capabilities. It is supported on ServerToAgent and AgentToServer. When this message
+is sent, the list of supported custom capabilities should be updated to match this list.
+If this message is never sent, it is assumed that no custom message types are supported.
+
+Capabilities are identified by a reverse FQDN with optional version information. For
+example, "com.company.capability/v2" identifies version 2 of "capability" created by
```
tigrannajaryan, line 2562: Is there a defined format for version number, e.g. semver?

proto/opamp.proto
andykellr: Yes. It makes handling the messages easier because you can switch on the type and decode into a struct that corresponds to the message type.

proto/opamp.proto
andykellr: The intent is that this is only sent when it changes, but we could also add another flag for `ReportCapabilities` for both the ServerToAgent and AgentToServer messages.

proto/opamp.proto
andykellr: A capability describes a custom feature that is supported and several custom messages may be involved in implementing that feature. For example, `io.opentelemetry.pause` which is described in the `Examples` section uses `pause`, `status`, and `error` messages. The message types are specific to the capability and the number of messages required and the direction of the messages is defined by the capability.

specification.md
andykellr: This isn't used at all by the protocol but we wanted to encourage implementers to follow a standard convention. I think semver is fine and we could document it as such.

proto/opamp.proto
tigrannajaryan: >The intent is that this is only sent when it changes

This is fine, let's make sure it is documented. I don't see it mentioned in the spec.

proto/opamp.proto
```diff
@@ -817,3 +833,33 @@ message AgentConfigFile {
...
     string content_type = 2;
 }
+
+/////////////////////////////////////////////////////////////////////////////////////
+// Custom messages
+/////////////////////////////////////////////////////////////////////////////////////
+
+message CustomCapabilities {
+    // A list of custom capabilities that are supported. Each capability is a reverse FQDN
+    // with optional version information that uniquely identifies the custom capability
```
tigrannajaryan, line 843: >is a reverse FQDN with optional version information

This part seems underspecified. My understanding is that FQDN usage is what ensure that are no clashes between custom messages introduced by different agents. For this to work correctly we need to define which part of the string is the FQDN and which part is the message. I can see that we have an example that clarifies it, but I do not think just an example is sufficient. I think we also need some wording here that specifies the format of this string.

proto/opamp.proto
andykellr: I added another sentence to the CustomCapabilities section which should clarify this.

> CustomCapabilities

> The Agent and Server should both use this message to signal that they support specific custom capabilities. It is supported on ServerToAgent and AgentToServer. This message should only be sent when the list of supported custom capabilities changes. When this message is sent, the list of supported custom capabilities should be updated to match this list. If this message is never sent, it is assumed that no custom capabilities are
supported.

specification.md
```diff
@@ -2474,6 +2531,217 @@ authenticity of the downloaded file, for example can be the
...
+Servers. Where appropriate, implementers are encouraged to document their use of
+CustomMessage so that support can be added to additional platforms. If appropriate, custom
+capabilities that are widely supported by Server and Agent implementations may be adopted
+as standard OpAMP capabilities in the future.
+
+#### CustomCapabilities
+
+The Agent and Server should both use this message to signal that they support specific
+custom capabilities. It is supported on ServerToAgent and AgentToServer. This message
+should only be sent when the list of supported custom capabilities changes. When this
```
tigrannajaryan, line 2560: >This message should only be sent when the list of supported custom capabilities changes.

👍 

When using an http transport it is possible for the server to observe capabilities change if the agent is updated to a new version and restarted. So this really can happen.

Note that if `ReportFullState` flag is used by the Server the Agent must report the full list of capabilities.

Can we mention `CustomCapabilities` in the "Agent Status Compression" section to make this clear?

proto/opamp.proto
andykellr: Let me know if you think the new language resolves this.

specification.md
```diff
@@ -2474,6 +2532,218 @@ authenticity of the downloaded file, for example can be the
...
+message is sent, the list of supported custom capabilities should be updated to match this
+list. If this message is never sent, it is assumed that no custom capabilities are
+supported.
+
+Capabilities must be identified by a reverse FQDN with optional version information. This
+avoids collisions between capabilities introduced by different organizations. The
+capability name is composed of the reverse FQDN of the organization that created the
+custom capability, a period, and a short name identifying the capability. The version is
+optional and is separated from the capability name by a forward slash, begins with the
+letter "v", and is followed by a number identifying the version.
```
tigrannajaryan, line 2572: Do we need the "v" part and do we need to say that the version is a number? Since we don't interpret it in any way perhaps we simply say the forward slash is follow by version number, e.g.

```suggestion
optional and is separated from the capability name by a forward slash.
```

specification.md
```diff
@@ -2474,6 +2532,217 @@ authenticity of the downloaded file, for example can be the
...
+data for each custom message type it defines.
+
+#### Examples
+
+The following examples describe possible uses of CustomCapabilities and CustomMessage but
+are not intended to be part of any specification. In these examples, "io.opentelemetry."
+is used to create the FQDN of the message types but each vendor is expected to use their
+own FQDN.
+
+##### Pause/Resume
```
tigrannajaryan, line 2635: Nit: just to make sure it is clearly an example:
```suggestion
##### Pause/Resume Example
```

specification.md
```diff
@@ -2474,6 +2532,217 @@ authenticity of the downloaded file, for example can be the
...
+"resume" and on success the new status would be `{ "paused": false }`.
+
+```
+CustomMessage {
+  capability: "io.opentelemetry.pause"
+  type: "resume"
+}
+```
+
+##### Service Discovery
```
tigrannajaryan, line 2704:
```suggestion
##### Service Discovery Example
```

tigrannajaryan (approved):
LGTM.

Just 2 small nits and we can merge after that.

Issue #133, opened 2022-10-03
Consider adding a header for WebSocket messages to allow future extensibility


tigrannajaryan:
## Problem

The WebSocket messages currently are serialized Protobuf AgentToServer/ServerToAgent messages. While Protobuf allows to easily add new fields to AgentToServer/ServerToAgent messages, there is no easy way to make any changes to how the Protobuf messages are serialized and stored in WebSocket messages. We will be essentially locked into this design decision.

## Proposed Solution

I would like to consider adding a simple extensibility mechanism to this part of the protocol - a one-byte header for WebSocket messages. Every WebSocket message is composed of 2 parts: one-byte header, followed by the serialized Protobuf AgentToServer/ServerToAgent messages. 

The current value of the one-byte header is equal to 0. In the future when non-zero values of headers are introduced their use should be negotiated during the connection phase (via HTTP headers or via Capabilities fields). This is necessary to ensure interoperability between different OpAMP versions. Non-zero header values will not be a breaking change, they will only used after successful negotiation between the Client and Server.

Note: we don't need the one-byte header for plain HTTP transport. Any extensibility for HTTP can be done via HTTP headers (not suitable for WebSocket since it cannot work per message - HTTP headers are sent once per connection).

## Example Usage

The one-byte header can be used in the future for example in the following ways.

### Message Fragmentation

We found that some WebSocket implementations (e.g. [AWS API Gateway](https://github.com/open-telemetry/opamp-spec/issues/60#issuecomment-1252659494)) limit the message size to 128KB. The only way to overcome this limitation is to perform message fragmentation and reassembly in OpAMP itself. This can be done by using one bit in the header to indicate whether the message is the last fragment.

### Support Other Compression Algorithms

WebSocket has built-in compression. What compression is used depends on the WebSocket implementation and realistically only "deflate" compression implementations are available. There are more modern compression algorithms (e.g. zstd), but it is impossible to use in OpAMP since most known WebSocket implementations simply don't support them.

The one-byte header can easily allow supporting these other compression algorithms. We can use one or or more of the header bits to indicate the compression algorithm used.

Note: this sort of extension can be also done via HTTP header negotiation. However, not all WebSocket implementations allow custom HTTP headers (browsers don't).

## Is This a Breaking Change?

It is, but we can handle it gracefully for a period of time. Here is how.

[...]

Issue #60, opened 2022-01-28
Test with client and server having different frame size limits


tigrannajaryan:
I was advised that different websocket implementations may have different frame size limits (typically 32K or64K bytes) which may result in interoperability problems. This needs testing and confirming that either it is not an issue or it needs a solution/workaround.

andykellr:
While investigating this issue, many of the references I found regarding frame size limits refer to the 32K limit in AWS API Gateway.

> Because of the WebSocket frame-size quota of 32 KB, a message larger than 32 KB must be split into multiple frames, each 32 KB or smaller. If a larger message (or larger frame size) is received, the connection is closed with code 1009.

https://docs.aws.amazon.com/apigateway/latest/developerguide/limits.html

This particular problem would present itself if an OpAMP server were hosted in AWS behind API Gateway and Agents were connecting to the server using an OpAMP client implementation with a WebSocket library that sent frames larger than 32K.

tigrannajaryan:
Thanks for the research.

What do you think if we add a (SHOULD) recommendation that Client implementations must limit outgoing message size to 32KB?

I wonder what frame size our current opamp-go implementation uses.

tigrannajaryan:
Apparently it is not me who thinks 32KB frame/128KB message is unreasonably low limit: https://repost.aws/questions/QUpz6uWU0SSnunjzPrbmLtRg/api-gateway-websockets-size-limit-feedback

tigrannajaryan:
I don't think we want to do message fragmentation and reassembly ourselves. That would dramatically complicate OpAMP just to accommodate one particular implementation that has unusually low limits.

I think we can probably do is this:
- Add a warning in the spec that some implementations have low frame/message size limits that may cause delivery failures and recommend to use plain HTTP transport if that happens.
- In our opamp-go implementation try to fragment large messages into small frames. Perhaps the frame size can be a client/server config option. For Gorilla Websockets using [NextWriter](https://pkg.go.dev/github.com/gorilla/websocket?utm_source=godoc#Conn.NextWriter) should result in slicing into frames of at most WriteBufferSize. 
- We should probably also file a feature request to Gorilla Websockets to add frame size limit as a proper option so that we don't rely on an undocumented behavior. 

Any other ideas?

tigrannajaryan:
File a feature request to gorilla websocket for now: https://github.com/gorilla/websocket/issues/814

tigrannajaryan:
[...]

PR #138, opened 2022-11-09
Adding a header for WebSocket messages to allow future extensibility

tigrannajaryan:
Resolves https://github.com/open-telemetry/opamp-spec/issues/133

Thi is ane extensibility mechanism - a header for WebSocket messages. Every WebSocket message is composed of 2 parts: varint-encoded unsigned 64 bit integer header, followed by the serialized Protobuf AgentToServer/ServerToAgent messages.

The current value of the header is equal to 0. In the future when non-zero values of headers are introduced their use should be negotiated during the connection phase (via HTTP headers or via Capabilities fields). This is necessary to ensure interoperability between different OpAMP versions. Non-zero header values will not be a breaking change, they will only used after successful negotiation between the Client and Server.

Note: we don't need the header for plain HTTP transport. Any extensibility for HTTP can be done via HTTP headers (not suitable for WebSocket since it cannot work per message - HTTP headers are sent once per connection).

## Example Usage

The header can be used in the future for example in the following ways.

### Message Fragmentation

We found that some WebSocket implementations (e.g. [AWS API Gateway](https://github.com/open-telemetry/opamp-spec/issues/60#issuecomment-1252659494)) limit the message size to 128KB. The only way to overcome this limitation is to perform message fragmentation and reassembly in OpAMP itself. This can be done by using one bit in the header to indicate whether the message is the last fragment.

### Support Other Compression Algorithms

WebSocket has built-in compression. What compression is used depends on the WebSocket implementation and realistically only "deflate" compression implementations are available. There are more modern compression algorithms (e.g. zstd), but it is impossible to use in OpAMP since most known WebSocket implementations simply don't support them.

The header can easily allow supporting these other compression algorithms. We can use one or more of the header bits to indicate the compression algorithm used.

Note: this sort of extension can be also done via HTTP header negotiation. However, not all WebSocket implementations allow custom HTTP headers (browsers don't).

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -12,6 +12,8 @@ Note: this document requires a simplification pass to reduce the scope, size and
 - [Introduction](#introduction)
 - [Communication Model](#communication-model)
   * [WebSocket Transport](#websocket-transport)
+    + [WebSocket Message Format](#websocket-message-format)
+    + [WebSocket Message Exchange](#websocket-message-exchange)
   * [Plain HTTP Transport](#plain-http-transport)
   * [AgentToServer and ServerToAgent Messages](#agenttoserver-and-servertoagent-messages)
     + [AgentToServer Message](#agenttoserver-message)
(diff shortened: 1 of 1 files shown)


specification.md
```diff
@@ -288,22 +290,68 @@ Supervisor is part of that entity.
...
+compliant with this specification SHOULD check that the value of the `header` is equal
+to 0 and if it is not SHOULD assume that the WebSocket message is malformed.
+
+The `data` field contains the bytes that represent the AgentToServer or ServerToAgent
+message encoded in [Protobuf binary wire format](
+https://developers.google.com/protocol-buffers/docs/encoding).
+
+Note that both `header` and `data` fields contain a variable number of bytes.
+The decoding Base 18 Varint algorithm for the `header` knows when to stop based on the
+bytes it reads.
```
andykellr, line 342:
```suggestion
Note that both `header` and `data` fields contain a variable number of bytes.
The decoding Base 128 Varint algorithm for the `header` knows when to stop based on the
bytes it reads.
```

andykellr (changes requested):
Typo, otherwise looks good.

specification.md

andykellr (approved):

Issue #134, opened 2022-10-04
Explain what "Stable" label means for OpAMP


tigrannajaryan:
We need a precise definition of stability, what exactly we promise by labeling the protocol "Stable".

tigrannajaryan:
We can likely borrow the definition from OTLP once it is ready: https://github.com/open-telemetry/opentelemetry-proto/issues/400

PR #141, opened 2023-01-05
Define OpAMP Protobuf schema stability guarantees

tigrannajaryan:
The guarantees mostly mirror the guarantees for OTLP (https://github.com/open-telemetry/opentelemetry-proto/pull/432), the primary difference being that OpAMP has no gRPC services defined and does not need any guarantees for service definitions.

This PR reflects the OpAMP Workgroup's discussions.

Resolves https://github.com/open-telemetry/opamp-spec/issues/134

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -154,6 +154,7 @@ Note: this document requires a simplification pass to reduce the scope, size and
   * [Interoperability of Future Capabilities](#interoperability-of-future-capabilities)
     + [Ignorable Capability Extensions](#ignorable-capability-extensions)
     + [Non-Ignorable Capability Extensions](#non-ignorable-capability-extensions)
+    + [Protobuf Schema Stability](#protobuf-schema-stability)
 - [Performance and Scale](#performance-and-scale)
 - [FAQ for Reviewers](#faq-for-reviewers)
   * [What is WebSocket?](#what-is-websocket)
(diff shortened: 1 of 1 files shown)


tigrannajaryan:
@andykellr I am opening this draft so that we can discuss what else is necessary to resolve https://github.com/open-telemetry/opamp-spec/issues/134

andykellr:
I think this looks good.

specification.md
```diff
@@ -2571,6 +2572,30 @@ peer and adjust their behavior appropriately. How exactly the behavior is
...
+[Protobuf definitions](proto/opamp.proto) for OpAMP 1.0:
+
+- Field types, numbers and names will not change.
+- Names of messages and enums will not change.
+- Names of enum choices and numbers assigned to enum choices will not change.
+- The location of messages and enums, i.e. whether they are declared at the top lexical
+  scope or nested inside another message will not change.
+- Package names and directory structure will not change.
+- Existing symbols will not be deleted.
+- `optional` and `repeated` declarators of existing fields will not change.
```
PeterF778, line 2587: Theoretically, since adding new fields is allowed in the future, perhaps we should not guarantee preservation of `optional` declarators.

tigrannajaryan: The reason we want to preserve the `optional` declarator is because changing can change the generated code (see [Protobuf presence API](https://github.com/protocolbuffers/protobuf/blob/main/docs/field_presence.md#presence-in-proto3-apis)). This can then result in breaking the OpAMP language library's API if the generated code is part of the API (which it is in opamp-go, for example).

PeterF778: Got it. Dropping `optional` results in dropping hazzers (in proto3) in the generated code.
BTW, I noticed that the current proto files for OpAMP do not use the `optional` label, but sometimes mention optionality in comments. Has this been done on purpose?

tigrannajaryan: > BTW, I noticed that the current proto files for OpAMP do not use the `optional` label, but sometimes mention optionality in comments. Has this been done on purpose?

The support of presence functions for `optional` fields was added relatively recently and the support in Protobuf generators was not very good last time I checked (some languages support it some others don't), so it was safer not to use `optional` and rely on comments at the time. The situation may be different now.

You are right though that we don't use `optional` currently and _adding_ it to any field should be backward compatible (_removing_ is not backward compatible though). I think it is probably not a bad idea to keep this requirement, otherwise _optional_ may be added to a field and then remove in some other future version and end up breaking code.

This is probably stronger guarantee than we need, but it is benign since it doesn't restrict us in any way.

specification.md
```diff
@@ -2571,6 +2572,31 @@ peer and adjust their behavior appropriately. How exactly the behavior is
...
+- Numbers assigned to enum choices will not change.
+- Names of enum choices will not change.
+- The location of messages and enums, i.e. whether they are declared at the top lexical
+  scope or nested inside another message will not change.
+- Package names and directory structure will not change.
+- `optional` and `repeated` declarators of existing fields will not change.
+- No existing symbol will be deleted.
+
+Future versions of the OpAMP specification may be extended by modifying the
+Protobuf schema defined in this specification version. The following Protobuf schema
```
jaronoff97, line 2591: should we specify how deprecations will be handled for the spec?

tigrannajaryan: I haven't put anything thought into the subject of deprecations. I also don't see anything in OTLP spec to use as an inspiration. It is probably worth thinking about, but probably should be a separate PR.

tigrannajaryan:
@andykellr please review.

andykellr (approved):

Issue #146, opened 2023-06-13
Declare stability per capability or for all spec?


tigrannajaryan:
We discussed that we want to wait for some OpAMP implementations to complete before we declare the spec stable, so that there is time to learn from the implementations and adjust the spec if needed. However the implementations that are in progress don't utilize all OpAMP capabilities, so we can't expect to learn much about these areas.

Should we declare stability of the spec for the capabilities that we know are relatively well-tested and move the other capabilities to a separate section and mark as "Beta" addition to the spec for now?

The spec sections and capabilities I am less confident about and which I think are not used in any implementations so far are:
- [Connection Settings Management](https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#connection-settings-management)
- [Packages](https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#packages) and [PackageStatuses Message](https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#packagestatuses-message)

We can probably mark these sections as `[Beta]` and the rest of the sections can be declared `Stable`.

@open-telemetry/opamp-spec-approvers what do you think?

andykellr:
I think that makes sense. BindPlane does use Packages and PackageStatuses for agent upgrade and it was worked well. It does not currently use ConnectionSettings.

tigrannajaryan:
> I think that makes sense. BindPlane does use Packages and PackageStatuses for agent upgrade and it was worked well. It does not currently use ConnectionSettings.

@andykellr are there any other capabilities that BindPlane does not use? Do you utilize everything except Connection Settings?

andykellr:
> > I think that makes sense. BindPlane does use Packages and PackageStatuses for agent upgrade and it was worked well. It does not currently use ConnectionSettings.
> 
> @andykellr are there any other capabilities that BindPlane does not use? Do you utilize everything except Connection Settings?

ConnectionSettings and AgentIdentification are not currently used by BindPlane.

tigrannajaryan:
Discussed in workgroup meeting today and decided that it is reasonable to declare only portions of the spec stable and keep less-tested parts as beta. I will work on figuring out the exact labeling mechanism.

tigrannajaryan:
Done in https://github.com/open-telemetry/opamp-spec/pull/147

PR #147, opened 2023-06-15
Mark certain capabilities as Beta

tigrannajaryan:
Contributes to https://github.com/open-telemetry/opamp-spec/issues/146

As agreed we will mark certain capabilities as Beta for now, until we have more confidence that they are well-specified. This will be primarily driven by feedback from implementations.

Capabilities marked Beta:
- Packages
- Connection Settings Managements
- Own Telemetry Reporting


tigrannajaryan:
@andykellr please take a look.

specification.md
```diff
@@ -524,9 +526,11 @@ enum AgentCapabilities {
     ReportsOwnLogs                 = 0x00000080;
```
srikanthccv, line 526: I see `ReportsOwn*` marked beta in proto but not here, is this intended?

srikanthccv (commented):
Small question about OwnTelemetry status inconsistency. LGTM otherwise!

specification.md
tigrannajaryan: Fixed.

srikanthccv (approved):

tigrannajaryan:
I added "Server to Agent Command" to the beta list.

tigrannajaryan:
@andykellr please take a look.

andykellr (approved):
I'm fine with these designations, but BindPlane is using the various Package messages and it is working well. I would be ok removing `Beta` from those capabilities.

Issue #165, opened 2023-08-17
Extend AgentHealth message to accommodate component health


mwear:
The addition of [component status reporting](https://github.com/open-telemetry/opentelemetry-collector/issues/7682) to the collector will allow us to have a more nuanced view of collector health by having visibility into the health of individual components. At a recent SIG meeting we discussed the possibility of extending the `AgentHealth` message to include a map that can be used to express the health of multiple components. There are open questions as to how we want restructure the message and how we can ensure that it remains agent agnostic. Do we want to keep the currently defined `AgentHealth` message for overall health, and add a map for component health, or do we want express everything as part of the map?

I'll suggest the following as a starting point, but expect that it is probably too coupled to our ideas around collector health. We can refine this as needed. This also retains the current `AgentHealth` fields as overall health.

```protobuf
message AgentHealth {
    bool healthy = 1;
    fixed64 start_time_unix_nano = 2;
    string last_error = 3;
    map<string, ComponentHealth> component_health_map = 4;
}

message ComponentHealth {
    fixed64 start_time_unix_nano = 1;
    ComponentStatusEnum status = 2;
    string last_error = 3;
}

enum ComponentStatusEnum {
    ComponentStatus_Starting = 0;
    ComponentStatus_OK = 1;
    ComponentStatus_RecoverableError = 2;
    ComponentStatus_PermanentError = 3;
    ComponentStatus_FatalError = 4;
    ComponentStatus_Stopping = 5;
    ComponentStatus_Stopped = 6;
}
```

I suspect dropping the status enum and replacing it with a string would be a good middle ground for making the ComponentHealth message more generalizable for non-collector agents. If anyone has thoughts or opinions, please comment.

tigrannajaryan:
Discussed in WG meeting today. I believe @mwear plans take a look at the following:

- Maybe do recursive components?
- Move from enum to string for status
- Add `healthy` field per component?
- How/do we need to compress health status on the network?
- Involve operator contributors, see how components apply there.

mwear:
Based on the discussion at the last WG meeting, I have experimented with two additional ways to model the `AgentHealthMessage`. One is a simplified version of what was previously proposed by replacing the enum with a string and the other is a recursive message.

### Option 1: Enum to String

```protobuf
message AgentHealth {
    bool healthy = 1;
    fixed64 start_time_unix_nano = 2;
    string last_error = 3;
    map<string, ComponentHealth> component_health_map = 4;
}

[...]

PR #168, opened 2023-10-06
Add ComponentHealth message

mwear:
As discussed in #165, this PR replaces the AgentHealth message with a wire-backwards-compatible ComponentHealth message. The ComponentHealth message allows reporting health for nested sub-components, component status, and status timestamp, while retaining the fields from the original AgentHealth message.

### DIFF
--- a/proto/opamp.proto
+++ b/proto/opamp.proto
@@ -46,8 +46,10 @@ message AgentToServer {
     // This field MUST be always set.
     uint64 capabilities = 4;
 
-    // The current health of the Agent.
-    // May be omitted if nothing changed since last AgentToServer message.
+    // The current health of the Agent and sub-components. The top-level ComponentHealth represents
+    // the health of the Agent overall. May be omitted if nothing changed since last AgentToServer
+    // message.
+    // Status: [Beta]
     ComponentHealth health = 5;
 
     // The current effective configuration of the Agent. The effective configuration is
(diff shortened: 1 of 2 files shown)


andykellr (approved):

evan-bradley (approved):

proto/opamp.proto
```diff
@@ -48,7 +48,7 @@ message AgentToServer {
 
     // The current health of the Agent.
     // May be omitted if nothing changed since last AgentToServer message.
-    AgentHealth health = 5;
+    ComponentHealth health = 5;
```
tigrannajaryan, line 51: Do we need to say that the top-level component represents the Agent as a whole?

proto/opamp.proto
```diff
@@ -621,19 +621,31 @@ enum AgentCapabilities {
     // Add new capabilities here, continuing with the least significant unused bit.
 }
 
-// The health of the Agent.
-message AgentHealth {
-    // Set to true if the Agent is up and healthy.
+// The health of the Agent and sub-components
```
tigrannajaryan, line 624: Should we mark this with "// Status: [Beta]" since it was just changed? That way when we declare 1.0 we still have a chance to fix this message in case we made any mistakes here.

proto/opamp.proto
mwear: I added a clarification. Let me know if it needs anything further.

tigrannajaryan (approved):
LGTM

PR #176, opened 2023-12-01
Add a field to specify keepalive interval to OpAMPConnectionSettings

### LINKED ISSUE #28: Add recommendations for keeping connections alive
1. Intermediaries such as load balancers may time out inactive connections. Add recommendations on how to keep it alive (either use WebSocket ping/pong or send empty status report).
2. Explain how clients should react to disconnections. Do they re-connection immediately? When do they pause (throttling) before re-connecting?

haoqixu:
Add an optional `keepalive_interval_seconds` to `OpAMPConnectionSettings` allowing OpAMP servers to adjust the keepalive interval.

Resolve #28


haoqixu:
see #190

PR #186, opened 2024-04-17
Replace ULIDs by 16 byte ids and recommend UUID v7

tigrannajaryan:
The early version of the spec was written before UUID v7 existed and ULID was chosen as a ID type that is suited to be used as a primary key in databases. Since than UUID v7 was standardized and aims to serve the exact same niche.

## Proposal

Replace instance_id definition to be an arbitrary sequence of 16 bytes. Both ULID and any version of UUID may be used as the value. We also recommend to use UUID v7 for the value.

This is a breaking change in the spec. We recommend opamp-go implementation to implement the change in backward compatible way, supporting both old and new instance_uid formats for a period of time.

Since `bytes` and `string` are compatible in the Protobuf wire encoding it is possible to use the new Protobuf declarations to receive messages encoded using the old Protobuf declarations - string's UTF8 bytes will simply populate the bytes of the new field. Receivers can simply probe for the length of the received bytes field and if it is 26 treat it as ULID in canonical format, if it is 16 bytes treat it as opaque id in new format, otherwise consider it invalid.

I will also post a PR in opamp-go that implements the above behavior.


andykellr (approved):

tigrannajaryan:
Corresponding implementation in opamp-go: https://github.com/open-telemetry/opamp-go/pull/272

tigrannajaryan:
@andykellr I am going to make a 0.9.0 release of the spec first, so that this change if merged goes into the next 0.10.0 release.

proto/opamp.proto
```diff
@@ -25,8 +25,8 @@ option go_package = "github.com/open-telemetry/opamp-go/protobufs";
 message AgentToServer {
     // Globally unique identifier of the running instance of the Agent. SHOULD remain
     // unchanged for the lifetime of the Agent process.
-    // Recommended format: https://github.com/ulid/spec
-    string instance_uid = 1;
+    // MUST be 16 bytes long and SHOULD be generated using UUID v7 spec.
```
evan-bradley, line 28:
```suggestion
    // MUST be 16 bytes long and SHOULD be generated using the UUID v7 spec.
```
Small nit. Same for the other mentions of the UUID v7 spec.

proto/opamp.proto
```diff
@@ -804,7 +804,8 @@ enum PackageStatusEnum {
 message AgentIdentification {
     // When new_instance_uid is set, Agent MUST update instance_uid
     // to the value provided and use it for all further communication.
-    string new_instance_uid = 1;
+    // MUST be 16 bytes long and SHOULD be generated using UUID v7 spec.
```
evan-bradley, line 807:
```suggestion
    // MUST be 16 bytes long and SHOULD be generated using the UUID v7 spec.
```

specification.md
```diff
@@ -504,9 +504,9 @@ The instance_uid field is a globally unique identifier of the running instance
 of the Agent. The Agent SHOULD self-generate this identifier and make the best
 effort to avoid creating an identifier that may conflict with identifiers
 created by other Agents. The instance_uid SHOULD remain unchanged for the
-lifetime of the Agent process. The instance_uid MUST be a
-[ULID](https://github.com/ulid/spec) formatted as a 26 character string in canonical
-representation.
+lifetime of the Agent process. The instance_uid MUST be
+16 bytes long and SHOULD be generated using
```
evan-bradley, line 508:
```suggestion
16 bytes long and SHOULD be generated using the
```

specification.md
```diff
@@ -826,12 +826,12 @@ enum ServerCapabilities {
 Properties related to identification of the Agent, which can be overridden by the
 Server if needed. When new_instance_uid is set, Agent MUST update instance_uid
 to the value provided and use it for all further communication. The new_instance_uid MUST
-be a [ULID](https://github.com/ulid/spec) formatted as a 26 character string in canonical
-representation.
+be 16 bytes long and SHOULD be generated using
```
evan-bradley, line 829:
```suggestion
be 16 bytes long and SHOULD be generated using the
```

evan-bradley (approved):

tigrannajaryan:
@andykellr I want to keep opamp-go examples interoperable with Bindplane OP. If we go ahead with this and then also merge https://github.com/open-telemetry/opamp-go/pull/272 do you think you will be able to quickly make Bindplane compatible with the new spec so that opamp-go client example can work with it?

proto/opamp.proto
tigrannajaryan: Fixed.

proto/opamp.proto
tigrannajaryan: Fixed.

specification.md

specification.md

proto/opamp.proto
```diff
@@ -25,8 +25,8 @@ option go_package = "github.com/open-telemetry/opamp-go/protobufs";
 message AgentToServer {
     // Globally unique identifier of the running instance of the Agent. SHOULD remain
     // unchanged for the lifetime of the Agent process.
-    // Recommended format: https://github.com/ulid/spec
-    string instance_uid = 1;
+    // MUST be 16 bytes long and SHOULD be generated using the UUID v7 spec.
+    bytes instance_uid = 1;
```
jaronoff97, line 29: this change opens up the instance id to contain a string that has non-utf8 encoded characters. Is that something we need to worry about here? I think the answer is no, because we actually do want a byte array to represent a UUID.

jaronoff97 (approved):
one quick question

proto/opamp.proto
tigrannajaryan: Yes, that's intentional. We want arbitrary bytes to be valid, not just UTF8 sequences.

jaronoff97: yep assumed as much, thanks for confirming.

tigrannajaryan:
> > @andykellr I want to keep opamp-go examples interoperable with Bindplane OP. If we go ahead with this and then also merge [open-telemetry/opamp-go#272](https://github.com/open-telemetry/opamp-go/pull/272) do you think you will be able to quickly make Bindplane compatible with the new spec so that opamp-go client example can work with it?
> 
> Bindplane only requires a string.

@andykellr Will it work correctly if the string contains arbitrary bytes? Protobuf spec says strings must be UTF8 compliant although not all receiver implementations validate it.

tigrannajaryan:
> Interesting. It sounds like this will present a problem until we expect the id to be an arbitrary array of 16 bytes.

@andykellr I am not sure Go protobuf deserializer validates this. It may be worth checking. If there is no validation it will be seen as just another string (possible with invalid UTF8 sequences).

tigrannajaryan:
> I did check this against BindPlane, looks like the deserializer does indeed validate that the string is UTF-8. It'll definitely be a priority for us to update after this and [open-telemetry/opamp-go#272](https://github.com/open-telemetry/opamp-go/pull/272) is merged; I think our goal would be to have BindPlane updated to support a week or 2 after it's released.

@BinaryFissionGames @andykellr OK, I think we are ready here in the spec and in opamp-go. If the timing works for you we can go ahead and merge this PR.

BinaryFissionGames:
@tigrannajaryan Yep, we're good with merging this 👍

PR #188, opened 2024-04-25
Move Dan Jaglowski to Emeritus Approver

djaglowski:
My participation in this SIG has been quite limited for some time. Correspondingly, I believe it would be appropriate to move to emeritus status at this time.


tigrannajaryan (approved):
Thanks @djaglowski !

Issue #183, opened 2024-04-05
OpAMP Agent heartbeats


jaronoff97:
Currently, if an agent is functioning successfully and without any changes to its health or component status, no messages will be sent after initial handshakes. For websockets, because no messages are sent from the agent to the server, after a certain period most of these connections will idle and be shut down causing a broken socket. I propose that we add support for an heartbeat interval as part of the specification. I have already implemented this capability [in the opamp-bridge](https://github.com/open-telemetry/opentelemetry-operator/blob/main/cmd/operator-opamp-bridge/agent/agent.go#L247-L267) where a user is able to configure a heartbeat interval to periodically set the agent's health message. HTTP theoretically gets around this with its polling interval.

This heartbeat implementation is important for the bridge where many of the events in Kubernetes happen asynchronously. The bridge is not informed directly therefore must poll state to send this message. The collector's opamp extension, however, watches for some changes but is currently not busy. For the extension, the collector will idle after sometime and prevent the server from being able to send any more messages to the extension. 

I think a heartbeat interval could be optional, however, it must be communicated as part of the initial AgentToServer message. This would allow the server to know when to mark the agent as unhealthy.

### Open Questions
* Should this functionality only exist for the socket transport?
  * I believe it would be valuable for this to exist regardless of the transport and that the HTTP poll interval could be deprecated in favor of this. This would allow the server to make decisions about the liveness of the agent regardless of the transport
* What should the default interval be? 30s?
* Should the heartbeat interval be negotiable from the server?

I'm happy to write the spec change for this issue, but would love everyone's thoughts here.

JaredTan95:
I think it is reasonable for the agent to actively report a healthy heartbeat, this helps opamp server keep the information up to date.

jaronoff97:
I think the main difference is that my proposal relates to heartbeats as a means of solving #28, allowing clients to automatically report health on an interval. A goal of this is definitely to keep the connection alive, but it also means that the client has a responsibility to reports its health on an interval. I can understand closing this issue in favor of that though...

haoqixu:
A crashed or misbehaving client may cause connection/goroutine leaks in the OpAMP server (open-telemetry/opamp-go#271). A heartbeat mechanism can help the server find out unresponsive peers and also defend against intermediaries (LB, proxy, network equipment) which may time out and terminate idle connections (#28).

haoqixu:
[...]

PR #190, opened 2024-05-07
Introduce heartbeats

jaronoff97:
This PR introduces a client side heartbeat that can be set via the OpAMP Connection Settings. After discussing more in the past SIG meeting, I worked through the following pro/con for a client side vs server side heartbeat.

closes https://github.com/open-telemetry/opamp-spec/issues/183

## Client Side
A client side heartbeat would be negotiated between the client and server, where the client's default heartbeat is set to 30s. On connect, the server has an option to entirely disable heartbeats by setting an explicit 0 for the field. The server can also offer a different heartbeat interval depending on its infrastructure's needs. After connection, the agent will begin a timer and each N seconds will send a message that minimally contains its instance id. An HTTP connection will use this interval for its polling as well. Some agents that are not directly informed of health changes should also use this for component health reporting. 

### Pros
1. The agent is able to prevent a proxy from timing out the socket connection
2. The agent's HTTP polling interval is now configurable
3. The server can properly age out and remove dead agents
4. Requires a single successful message for heartbeat processing

### Cons
1. requires a proto change
2. extra work for the client
  a. the client will now need to keep track of a heartbeat timer to send this periodic message

## Server Side
A server side heartbeat would simply be a part of opamp-go and would require no changes to the spec to allow this to work. The server every N seconds would send an empty message to the client to keep the socket connection active. 

### Pros
1. Requires no spec or proto changes
2. Server is in control of the interval

### Cons
1. Server has no way to determine if an Agent is dead
  a. The core value of the change to me is that the server can now rely on the fact it is receiving a message every N seconds and can take action if that is the case
  b. If the client is using an http transport, there is no way for the server to reliably send a heartbeat message to guarantee the liveness of the agent. Say the server 'requests' a heartbeat from the client, but the client is already dead 
2. The only way to send an 'empty' server to agent message today is by using the report full state flag. This means the message back from the agent is going to be larger than necessary solely to keep the connection.
  a. We could also add a heartbeat flag as part of the message
3. Requires three successful messages for heartbeat processing
  a. A server would need to successfully send the heartbeat flagged message over the websocket, the client would then send its heartbeat back via an AgentToServer message, and the server would need to ACK with a responding ServerToAgent message. 

I think given the pros and cons of the above, I prefer an Agent heartbeat over a server heartbeat. If we need a new proto change anyway to introduce a heartbeat flag, I think the client approach is more effective. Furthermore, this change helps provide guidance for agents that are not informed of status updates. By setting an explicit heartbeat for the client, the server can increase the granularity of the agent's status updates. The client heartbeat approach also matches the design for a conventional deadman's switch – something that is constantly sending a signal out for a receiver to detect only when that signal is no longer received. Flipping that design removes that guarantee and weakens the overall feature. 

Finally, the server would also be able to explicitly disconnect misbehaving clients and force them to reconnect with new settings. If the server were to not receive a heartbeat within its set window, the server could initiate a disconnect to gracefully close the client. This approach would not work as well for a server heartbeat as it would need to cancel the initial server to agent message.

## Open Questions
* Where should we put more information about client-side heartbeating should work

## References
- [RabbitMQ](https://www.rabbitmq.com/docs/heartbeats#tcp-keepalives)
  - Rabbit prefers client heartbeats over server ones, AND explicit heartbeat instead of TCP keepalives
- [MQTT](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901094)
  - Functionally the same as this proposal, they call them keep alives
- [Phoenix Socket Client](https://hexdocs.pm/phoenix/writing_a_channels_client.html#message-format)
  - Works by the client sending a specific heartbeat message


proto/opamp.proto
```diff
@@ -264,6 +264,20 @@ message OpAMPConnectionSettings {
...
     // This field can be used to perform a client certificate revocation/rotation.
     TLSCertificate certificate = 3;
+
+    // The Agent MUST periodically send an AgentToServer message if this field is set. At a minimum
+    // the instance_uid field MUST be set. It is recommended that the Agent also set ComponentHealth
+    // as well.
+    //
+    // A Polling-based HTTP Client MUST use the value as polling interval.
+    //
+    // A heartbeat is used to keep a load balancer connection active AND inform the server that the Agent
```
mwear, line 274: Does "AND" need to be in caps?
```suggestion
    // A heartbeat is used to keep a load balancer connection active and inform the server that the Agent
```

specification.md
```diff
@@ -1855,6 +1860,20 @@ for this connection.
...
 This field can be used to perform a client certificate revocation/rotation.
 
+##### OpAMPConnectionSettings.heartbeat_interval_seconds
+
+The Client should use the offered heartbeat interval to periodically send an AgentToServer
+message. At a minimum the instance_uid field MUST be set. It is recommended that the Agent
+also set ComponentHealth as well. An HTTP based-client MUST use the heartbeat interval as
+its polling interval.
+
+A heartbeat is used to keep a load balancer connection active AND inform the server that
```
mwear, line 1870: Same here, does "AND" need to be in caps?
```suggestion
A heartbeat is used to keep a load balancer connection active and inform the server that
```

mwear (commented):
Mostly LGTM. The reasoning behind client side heartbeats makes sense and your references to other well known projects helps validate that decision.

specification.md
jaronoff97: copied 😅

specification.md
```diff
@@ -220,19 +221,20 @@ mixed Agents from different vendors.
 
 OpAMP supports the following functionality:
 
-* Remote configuration of the Agents.
-* Status reporting. The protocol allows the Agent to report the properties of
+- Remote configuration of the Agents.
```
mwear, line 224: FYI, your editor changed `*` to `-` everywhere, which makes the diff larger than it should be.
```suggestion
* Remote configuration of the Agents.
```

jaronoff97: yeah :( i couldn't figure out how it decided what to change here... i can change it back if it really matters

mwear: I don't have strong opinions, but I generally try to leave the file how I found it.

specification.md
```diff
@@ -1855,6 +1860,20 @@ for this connection.
...
+##### OpAMPConnectionSettings.heartbeat_interval_seconds
+
+The Client should use the offered heartbeat interval to periodically send an AgentToServer
+message. At a minimum the instance_uid field MUST be set. It is recommended that the Agent
+also set ComponentHealth as well. An HTTP based-client MUST use the heartbeat interval as
+its polling interval.
+
+A heartbeat is used to keep a load balancer connection active and inform the server that
+the Agent is still alive and active. A server could use the heartbeat to make decisions about
+the liveness of the connected Agent.
```
BinaryFissionGames, line 1872: This is a breaking change as it's written, right? Agents pre-heartbeat will ignore the heartbeat interval, the server will determine they aren't alive, then (presumably) disconnect the agent.

I think this could easily be made backwards compatible if the agent explicitly indicated that it supports this e.g. through it's capabilities.

jaronoff97: maybe not, there's guidance [here](https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#ignorable-capability-extensions). I think given we also want to use this to control the HTTP polling interval this may be a non-ignorable capability extension and would therefore require a new AgentCapability. Is there anyway around that @tigrannajaryan @andykellr ?

andykellr: I think a new flag for `AgentCapabilities_ReportsHeartbeat` would be appropriate so that servers know if they should expect heartbeats.

proto/opamp.proto
```diff
@@ -264,6 +264,20 @@ message OpAMPConnectionSettings {
...
+    //
+    // A Polling-based HTTP Client MUST use the value as polling interval.
+    //
+    // A heartbeat is used to keep a load balancer connection active and inform the server that the Agent
+    // is still alive and active.
+    //
+    // This field is optional:
+    // if omitted, a default heartbeat interval of 30 seconds should be used.
+    // if set to zero, the heartbeat ability will be disabled.
+    optional uint64 heartbeat_interval_seconds = 4;
```
tigrannajaryan, line 280: IIRC, some Protobuf implementations can't differentiate between "not present" and zero values (e.g. Gogoproto). Can we have a different way that does not rely on this capability?

specification.md
```diff
@@ -444,9 +444,12 @@ deliver to the Agent (such as for example a new remote configuration).
 
 The default polling interval when the Agent does not have anything to deliver is 30
 seconds. This polling interval SHOULD be configurable on the Client.
+If the server sets OpAMPConnectionSettings.heartbeat_interval_seconds, the client MUST
```
tigrannajaryan, line 447: This section applies to http transport only. I think we need to add something in the websocket section too.

proto/opamp.proto
```diff
@@ -635,7 +649,10 @@ enum AgentCapabilities {
     AgentCapabilities_ReportsHealth                  = 0x00000800;
     // The Agent will report RemoteConfig status via AgentToServer.remote_config_status field.
     AgentCapabilities_ReportsRemoteConfig            = 0x00001000;
-
+    // The Agent can report heartbeats on a default interval of 30s.
```
tigrannajaryan, line 652:
```suggestion
    // The Agent will report heartbeats on a default interval of 30s.
```
I think "can" is not strong enough. It implies a possibility, but the server needs an assurance.

specification.md
```diff
@@ -1855,6 +1863,20 @@ for this connection.
...
+The Client should use the offered heartbeat interval to periodically send an AgentToServer
+message. At a minimum the instance_uid field MUST be set. It is recommended that the Agent
+also set ComponentHealth as well. An HTTP based-client MUST use the heartbeat interval as
+its polling interval.
+
+A heartbeat is used to keep a load balancer connection active and inform the server that
+the Agent is still alive and active. A server could use the heartbeat to make decisions about
+the liveness of the connected Agent.
+
+A default of a 30s should be used if not set by the OpAMPConnectionSettings. If the
```
tigrannajaryan, line 1877: A default of 30s second polling for WebSockets seems like an overkill. What was the typical timeout you were observing?

proto/opamp.proto
jaronoff97: yes, i can remove that bit now that I added heartbeat as part of the AgentCapabilities Enum

specification.md
jaronoff97: I've seen timeouts after 30s running locally and in cloud environments. Per [these AWS docs](https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-websocket-api-integration-requests.html) 30s appears to be their default.

jaronoff97: [GCP's is 5 mins ](https://cloud.google.com/run/docs/triggering/websockets#client-reconnects)

jaronoff97: [Azure recommends a 20-40 second timeout](https://learn.microsoft.com/en-us/azure/application-gateway/application-gateway-websocket#backendaddresspool-backendhttpsetting-and-routing-rule-configuration)

specification.md
```diff
@@ -1830,6 +1842,7 @@ message OpAMPConnectionSettings {
     string destination_endpoint = 1;
     Headers headers = 2;
     TLSCertificate certificate = 3;
+    optional uint64 heartbeat_interval_seconds = 4;
```
tpaschalis, line 1845: This is potentially an overkill/nit, but do you think we should use a [duration](https://protobuf.dev/reference/protobuf/google.protobuf/#duration) as the type?

We already have another field that is a duration but uses different units (`retry_after_nanoseconds`) which is potentially confusing. This is also in-line with [Google's AIPs](https://google.aip.dev/142).

specification.md
```diff
@@ -1855,6 +1868,19 @@ for this connection.
...
+If the ReportsHeartbeat capability is true, the Client MUST use the offered heartbeat
+interval to periodically send an AgentToServer message. At a minimum the instance_uid
+field MUST be set. It is recommended that the Agent also set ComponentHealth as well.
+An HTTP based-client MUST use the heartbeat interval as its polling interval.
+
+A heartbeat is used to keep a load balancer connection active and inform the server that
+the Agent is still alive and active. A server could use the heartbeat to make decisions about
+the liveness of the connected Agent.
+
+A default of a 30s should be used if not set by the OpAMPConnectionSettings.
```
tpaschalis, line 1882: As brought up in #183, should we write a note about clients respecting values negotiated from the server or coming from a backpressure mechanism (eg. a Retry-After header?)

tpaschalis (approved):
LGTM overall, and I think the client-side approach would work best.

jaronoff97:
@tigrannajaryan right now the bridge is not informed of the health of the agents it monitors, but instead uses the heartbeat interval to query the component health in the cluster. I figured that other agents that are doing similar interval-based health operations should do the same. I can clarify more in the spec to make that clear.

proto/opamp.proto
```diff
@@ -635,7 +648,10 @@ enum AgentCapabilities {
     AgentCapabilities_ReportsHealth                  = 0x00000800;
     // The Agent will report RemoteConfig status via AgentToServer.remote_config_status field.
     AgentCapabilities_ReportsRemoteConfig            = 0x00001000;
-
+    // The Agent will report heartbeats on a default interval of 30s.
+    // This is specified by the ServerToAgent.OpAMPConnectionSettings.heartbeat_interval_seconds field.
+    // Status: [Beta]
+    AgentCapabilities_ReportsHeartbeat               = 0x00002000;
```
tigrannajaryan, line 654: Can you please clarify what the purpose of this capability flag is? How is the server logic going to be different depending on this flag?

proto/opamp.proto
```diff
@@ -264,6 +264,19 @@ message OpAMPConnectionSettings {
     // This field is optional: if omitted the client SHOULD NOT use a client-side certificate.
     // This field can be used to perform a client certificate revocation/rotation.
     TLSCertificate certificate = 3;
+
+    // The Agent MUST periodically send an AgentToServer message if the
+    // AgentCapabilities_ReportsHeartbeat capability is true. At a minimum the instance_uid
+    // field MUST be set. It is recommended that the Agent also set ComponentHealth as well.
```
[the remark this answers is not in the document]
jaronoff97: yep that makes sense. I think individual clients can decide to send more if they want, and we don't need to be strong here.

proto/opamp.proto
jaronoff97: from past discussions we had decided that we were going to use the capability to determine whether or not heartbeats should be disabled. I don't think we should have multiple ways of disabling this.

proto/opamp.proto
jaronoff97: the server may set the `heartbeat_interval_seconds` field for the connection settings based on what the server knows its timeout may be.

proto/opamp.proto
tigrannajaryan: The operation is not entire clear to me. Who is making a decision to disable the heartbeats? Is it the Server or the Client?

proto/opamp.proto
jaronoff97: okay, can do.

proto/opamp.proto
jaronoff97: ah i see what you're saying. I think the flow is this:
1. client reports that it can send heartbeats
2. server sets a non-zero heartbeat interval that the client respects

With this in mind, either the client or the server can disable heartbeats at any time.

tigrannajaryan: OK. I think we need to describe the flows and possible options (with heartbeats, without heartbeats, client deciding, sever deciding, etc), what's the initial state, etc. 

It may be even worth showing sequence diagrams if plain text is not clear enough.

specification.md
```diff
@@ -1855,6 +1868,47 @@ for this connection.
...
+interval to periodically send an AgentToServer message. If the capability is true
+and the Server sets heartbeat_interval_seconds to 0, Agent heartbeats should be disabled.
+At a minimum the instance_uid field MUST be set.
+An HTTP based-client MUST use the heartbeat interval as its polling interval.
+
+A heartbeat is used to keep a load balancer connection active and inform the server that
+the Agent is still alive and active. A server could use the heartbeat to make decisions about
+the liveness of the connected Agent.
+
+The flow for negotiating a heartbeat is described as so:
```
mwear, line 1883:
```suggestion
The flow for negotiating a heartbeat is described as so:

```

mwear (approved):
LGTM. One formatting suggestion to appease markdown lint.

BinaryFissionGames (approved):
Looks good to me, thanks for pushing this forward!

proto/opamp.proto
```diff
@@ -635,7 +649,10 @@ enum AgentCapabilities {
     AgentCapabilities_ReportsHealth                  = 0x00000800;
     // The Agent will report RemoteConfig status via AgentToServer.remote_config_status field.
     AgentCapabilities_ReportsRemoteConfig            = 0x00001000;
-
+    // The Agent will report heartbeats on a default interval of 30s.
```
evan-bradley, line 652: Should we specify that 30 seconds is the default interval? Won't that vary by client implementation?

specification.md
```diff
@@ -1855,6 +1868,48 @@ for this connection.
 This field is optional: if omitted the client SHOULD NOT use a client-side certificate.
 This field can be used to perform a client certificate revocation/rotation.
 
+##### OpAMPConnectionSettings.heartbeat_interval_seconds
+
+If the ReportsHeartbeat capability is true, the Client MUST use the offered heartbeat
+interval to periodically send an AgentToServer message. If the capability is true
+and the Server sets heartbeat_interval_seconds to 0, Agent heartbeats should be disabled.
+At a minimum the instance_uid field MUST be set.
+An HTTP based-client MUST use the heartbeat interval as its polling interval.
```
evan-bradley, line 1877:
```suggestion
An HTTP-based client MUST use the heartbeat interval as its polling interval.
```
Nit

specification.md
```diff
@@ -1855,6 +1868,48 @@ for this connection.
...
+│          │                       │          │
+└──────────┘                       └──────────┘
+```
+
+1. The agent connects to the server and optionally sets the ReportsHeartbeat capability. If the Agent does NOT set this capability, no heartbeats will occur.
+2. If the Agent sets the ReportsHeartbeat capability, the server MUST respond with either a 0 indicating that heatbeats are disabled. Otherwise, the server will set a heartbeat_interval_seconds in the OpAMPConnectionSettings message. It is recommended for servers to set a 30s interval, if desired.
+3. If the Agent sets the ReportsHeartbeat capability AND the server has set OpAMPConnectionSettings.heartbeat_interval_seconds, the Agent MUST send a heartbeat message for the interval set by the server.
+4. The Agent will continue to send heartbeats on its configured interval while alive.
+
+The Agent can decide not to send heartbeats by not setting the ReportsHeartbeat capability. The Server can decide to not support heartbeats by responding with an unset (or 0) for the OpAMPConnectionSettings.heartbeat_interval_seconds.
```
evan-bradley, line 1911:
```suggestion
The Agent can decide not to send heartbeats by not setting the ReportsHeartbeat capability. The Server can decide to not support heartbeats by responding with a value of `0` seconds in the OpAMPConnectionSettings.heartbeat_interval_seconds field.
```
`unset` just means setting `0` here, right? `uint64` isn't an optional field, will that be translated to a special `unset` value if the server does not set it? Also, we say above that the server has to respond with a value, is there a way to mark it as unset?

specification.md
```diff
@@ -579,7 +587,10 @@ enum AgentCapabilities {
     ReportsHealth                  = 0x00000800;
     // The Agent will report RemoteConfig status via AgentToServer.remote_config_status field.
     ReportsRemoteConfig            = 0x00001000;
-
+    // The Agent can report heartbeats on a default interval of 30s.
+    // This is specified by the ServerToAgent.OpAMPConnectionSettings.heartbeat_interval_seconds field.
```
evan-bradley, line 591:
```suggestion
    // The heartbeat interval is specified by the ServerToAgent.OpAMPConnectionSettings.heartbeat_interval_seconds field.
```
Maybe a little redundant in wording, but I think it may help to specify what "this" is.

specification.md
```diff
@@ -1855,6 +1868,48 @@ for this connection.
...
+│          │      interval         │          │
+│          │                       │          │
+│          │ (4) Send Heartbeat    │          │
+│          ├──────────────────────►│          │
+│          │                       │          │
+└──────────┘                       └──────────┘
+```
+
+1. The agent connects to the server and optionally sets the ReportsHeartbeat capability. If the Agent does NOT set this capability, no heartbeats will occur.
+2. If the Agent sets the ReportsHeartbeat capability, the server MUST respond with either a 0 indicating that heatbeats are disabled. Otherwise, the server will set a heartbeat_interval_seconds in the OpAMPConnectionSettings message. It is recommended for servers to set a 30s interval, if desired.
```
evan-bradley, line 1907:
```suggestion
2. If the Agent sets the ReportsHeartbeat capability, the server MUST respond by setting an interval in the heartbeat_interval_seconds field within the OpAMPConnectionSettings message. The value can either the desired interval, or `0`, indicating that the client should not send heartbeats. 30s is the recommended default interval.
```
Feel free to update to your liking, I just thought the phrasing could use some tweaking.

evan-bradley (commented):
Looks good to me on whole, just a few notes regarding wording.

specification.md
jaronoff97: nope, this is a relic.

proto/opamp.proto
jaronoff97: In this case we are saying that by default a 30s interval should be used for heartbeating. Some servers may override this default.

specification.md
jaronoff97:
```suggestion
2. If the Agent sets the ReportsHeartbeat capability, the server MUST respond by setting an interval in the heartbeat_interval_seconds field within the OpAMPConnectionSettings message. The value can either be the desired interval, or `0`, indicating that the client should not send heartbeats. 30s is the recommended default interval.
```

specification.md
```diff
@@ -1855,6 +1868,48 @@ for this connection.
 This field is optional: if omitted the client SHOULD NOT use a client-side certificate.
 This field can be used to perform a client certificate revocation/rotation.
 
+##### OpAMPConnectionSettings.heartbeat_interval_seconds
+
+If the ReportsHeartbeat capability is true, the Client MUST use the offered heartbeat
+interval to periodically send an AgentToServer message. If the capability is true
+and the Server sets heartbeat_interval_seconds to 0, Agent heartbeats should be disabled.
+At a minimum the instance_uid field MUST be set.
```
tigrannajaryan, line 1876:
```suggestion

Any AgentToServer message where instance_uid field is set is considered a
valid heartbeat. Note that it is not necessary to send a separate AgentToServer
message just for heartbeating purposes if another AgentToServer message
containing other data was just sent. The Agent must count heartbeating interval
from the last AgentToServer message sent.

```

specification.md
```diff
@@ -2942,6 +2997,8 @@ response and MAY optionally set
 header to indicate when SHOULD the Client attempt to reconnect. The Client SHOULD
 honour the corresponding requirements of HTTP specification.
 
+Note: this reconnect is separate from [heartbeats](#opampconnectionsettingsheartbeat_interval_seconds). A client should still send regular heartbeat messages if it is configured to do so.
```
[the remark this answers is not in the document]
jaronoff97: no, the second sentence states that the client should continue sending heartbeats. I will clear up the wording.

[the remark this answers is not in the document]
tigrannajaryan: Is that desirable? If the server sends 503/429 it is probably overwhelmed. Sending heartbeats can make things worse. I think we should stop sending heartbeats for the duration specified in server's response.

proto/opamp.proto
```diff
@@ -635,7 +649,10 @@ enum AgentCapabilities {
...
+    // This is specified by the ServerToAgent.OpAMPConnectionSettings.heartbeat_interval_seconds field.
```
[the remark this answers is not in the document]
jaronoff97: I think that's reasonable. Adding.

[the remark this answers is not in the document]
tigrannajaryan: I had a typo, to be clear I meant:

>the Server may not know this and should not make assumptions about it

proto/opamp.proto
```diff
@@ -264,6 +264,19 @@ message OpAMPConnectionSettings {
...
+    // AgentCapabilities_ReportsHeartbeat capability is true. At a minimum the instance_uid
+    // field MUST be set.
+    //
+    // An HTTP Client MUST use the value as polling interval, if heartbeat_interval_seconds is non zero.
+    //
+    // A heartbeat is used to keep the connection active and inform the server that the Agent
+    // is still alive and active.
+    //
+    // If this field has no value or is set to 0,
+    // the Agent should not send any heartbeats.
```
evan-bradley, line 278:
```suggestion
    // If this field has no value or is set to 0, the Agent should not send any heartbeats.
```
Nit

proto/opamp.proto
```diff
@@ -264,6 +264,19 @@ message OpAMPConnectionSettings {
     // This field is optional: if omitted the client SHOULD NOT use a client-side certificate.
     // This field can be used to perform a client certificate revocation/rotation.
     TLSCertificate certificate = 3;
+
+    // The Agent MUST periodically send an AgentToServer message if the
+    // AgentCapabilities_ReportsHeartbeat capability is true. At a minimum the instance_uid
+    // field MUST be set.
+    //
+    // An HTTP Client MUST use the value as polling interval, if heartbeat_interval_seconds is non zero.
```
evan-bradley, line 272:
```suggestion
    // An HTTP Client MUST use the value as polling interval, if heartbeat_interval_seconds is non-zero.
```

proto/opamp.proto
```diff
@@ -635,7 +648,13 @@ enum AgentCapabilities {
     AgentCapabilities_ReportsHealth                  = 0x00000800;
     // The Agent will report RemoteConfig status via AgentToServer.remote_config_status field.
     AgentCapabilities_ReportsRemoteConfig            = 0x00001000;
-
+    // The Agent can report heartbeats.
+    // This is specified by the ServerToAgent.OpAMPConnectionSettings.heartbeat_interval_seconds field.
+    // If this capability is true, but the Server does not set a heartbeat_interval_seconds field, the
+    // Agent should use its own configured interval, which by default will be 30s, the Server may not know this
```
evan-bradley, line 654:
```suggestion
    // Agent should use its own configured interval, which by default should be 30s, the Server may not know this
```
We say "should" elsewhere, and I think we should leave it to client libraries to determine the default interval.

specification.md
```diff
@@ -579,7 +590,13 @@ enum AgentCapabilities {
     ReportsHealth                  = 0x00000800;
     // The Agent will report RemoteConfig status via AgentToServer.remote_config_status field.
     ReportsRemoteConfig            = 0x00001000;
-
+    // The Agent can report heartbeats.
+    // This is specified by the ServerToAgent.OpAMPConnectionSettings.heartbeat_interval_seconds field.
+    // If this capability is true, but the Server does not set a heartbeat_interval_seconds field, the
+    // Agent should use its own configured interval, which by default will be 30s, the Server may not know this
```
evan-bradley, line 596:
```suggestion
    // Agent should use its own configured interval, which by default should be 30s, the Server may not know this
```

specification.md
```diff
@@ -1855,6 +1874,54 @@ for this connection.
 This field is optional: if omitted the client SHOULD NOT use a client-side certificate.
 This field can be used to perform a client certificate revocation/rotation.
 
+##### OpAMPConnectionSettings.heartbeat_interval_seconds
+
+If the ReportsHeartbeat capability is true, the Client MUST use the offered heartbeat
+interval to periodically send an AgentToServer message. If the capability is true
+and the Server sets heartbeat_interval_seconds to 0, Agent heartbeats should be disabled.
```
evan-bradley, line 1881:
```suggestion
and the Server sets heartbeat_interval_seconds to 0, Agent heartbeats MUST be disabled.
```
I think if we require that the agent respects the interval set by the server, it needs to respect disabling heartbeats as well.

evan-bradley (approved):
LGTM, thanks for all your work on this.

proto/opamp.proto
```diff
@@ -635,7 +647,13 @@ enum AgentCapabilities {
     AgentCapabilities_ReportsHealth                  = 0x00000800;
     // The Agent will report RemoteConfig status via AgentToServer.remote_config_status field.
     AgentCapabilities_ReportsRemoteConfig            = 0x00001000;
-
+    // The Agent can report heartbeats.
+    // This is specified by the ServerToAgent.OpAMPConnectionSettings.heartbeat_interval_seconds field.
+    // If this capability is true, but the Server does not set a heartbeat_interval_seconds field, the
+    // Agent should use its own configured interval, which by default should be 30s, the Server may not know this
```
tigrannajaryan, line 653:
```suggestion
    // Agent should use its own configured interval, which by default should be 30s. The Server may not know the configured interval
```

specification.md
```diff
@@ -1855,6 +1874,54 @@ for this connection.
 This field is optional: if omitted the client SHOULD NOT use a client-side certificate.
 This field can be used to perform a client certificate revocation/rotation.
 
+##### OpAMPConnectionSettings.heartbeat_interval_seconds
+
+If the ReportsHeartbeat capability is true, the Client MUST use the offered heartbeat
+interval to periodically send an AgentToServer message. If the capability is true
+and the Server sets heartbeat_interval_seconds to 0, Agent heartbeats MUST be disabled.
+At a minimum the instance_uid field MUST be set.
```
tigrannajaryan, line 1882:
```suggestion
At a minimum the `AgentToServer.instance_uid` field MUST be set in the heartbeats.
```

specification.md
```diff
@@ -579,7 +590,13 @@ enum AgentCapabilities {
     ReportsHealth                  = 0x00000800;
     // The Agent will report RemoteConfig status via AgentToServer.remote_config_status field.
     ReportsRemoteConfig            = 0x00001000;
-
+    // The Agent can report heartbeats.
+    // This is specified by the ServerToAgent.OpAMPConnectionSettings.heartbeat_interval_seconds field.
+    // If this capability is true, but the Server does not set a heartbeat_interval_seconds field, the
+    // Agent should use its own configured interval, which by default should be 30s, the Server may not know this
```
tigrannajaryan, line 596:
```suggestion
    // Agent should use its own configured interval, which by default should be 30s, the Server may not know the configured interval
```

tigrannajaryan (approved):
LGTM, just a couple final nits.

Thanks for the patience @jaronoff97 !

tigrannajaryan:
@andykellr if you would like to take a look, otherwise I will merge it.

andykellr:
You can merge. I've been following along and looks great.

proto/opamp.proto
```diff
@@ -264,6 +264,18 @@ message OpAMPConnectionSettings {
...
+    // AgentCapabilities_ReportsHeartbeat capability is true. At a minimum the instance_uid
+    // field MUST be set.
+    //
+    // An HTTP Client MUST use the value as polling interval, if heartbeat_interval_seconds is non-zero.
+    //
+    // A heartbeat is used to keep the connection active and inform the server that the Agent
+    // is still alive and active.
+    //
+    // If this field has no value or is set to 0, the Agent should not send any heartbeats.
+    uint64 heartbeat_interval_seconds = 4;
```
tigrannajaryan, line 278:
```suggestion
    // Status: [Development]
    uint64 heartbeat_interval_seconds = 4;
```

proto/opamp.proto
```diff
@@ -635,7 +647,13 @@ enum AgentCapabilities {
     AgentCapabilities_ReportsHealth                  = 0x00000800;
     // The Agent will report RemoteConfig status via AgentToServer.remote_config_status field.
     AgentCapabilities_ReportsRemoteConfig            = 0x00001000;
-
+    // The Agent can report heartbeats.
+    // This is specified by the ServerToAgent.OpAMPConnectionSettings.heartbeat_interval_seconds field.
+    // If this capability is true, but the Server does not set a heartbeat_interval_seconds field, the
+    // Agent should use its own configured interval, which by default will be 30s. The Server may not
+    // know the configured interval and should not make assumptions about it.
+    // Status: [Beta]
```
tigrannajaryan, line 655:
```suggestion
    // Status: [Development]
```

specification.md
```diff
@@ -579,7 +590,13 @@ enum AgentCapabilities {
     ReportsHealth                  = 0x00000800;
     // The Agent will report RemoteConfig status via AgentToServer.remote_config_status field.
     ReportsRemoteConfig            = 0x00001000;
-
+    // The Agent can report heartbeats.
+    // This is specified by the ServerToAgent.OpAMPConnectionSettings.heartbeat_interval_seconds field.
+    // If this capability is true, but the Server does not set a heartbeat_interval_seconds field, the
+    // Agent should use its own configured interval, which by default will be 30s. The Server may not
+    // know the configured interval and should not make assumptions about it.
+    // Status: [Beta]
```
tigrannajaryan, line 598:
```suggestion
    // Status: [Development]
```

specification.md
```diff
@@ -1855,6 +1874,54 @@ for this connection.
 This field is optional: if omitted the client SHOULD NOT use a client-side certificate.
 This field can be used to perform a client certificate revocation/rotation.
 
+##### OpAMPConnectionSettings.heartbeat_interval_seconds
+
+If the ReportsHeartbeat capability is true, the Client MUST use the offered heartbeat
```
tigrannajaryan, line 1879:
```suggestion

Status: [Development]

If the ReportsHeartbeat capability is true, the Client MUST use the offered heartbeat
```

tigrannajaryan (commented):
One last thing I forgot to check earlier: the status labels.

Let's use `[Development]` for now.

johnphilip283:
great work, thank you @jaronoff97!

PR #192, opened 2024-05-22
Fix typos

jskiba:
Just some simple typos.
I'm not sure if I'm missing something and these descriptions should be somehow automatically copied to proto files or it had to be corrected manually in these two files (that's how I did it).


tigrannajaryan:
Thanks for the fixes.

> I'm not sure if I'm missing something and these descriptions should be somehow automatically copied to proto files or it had to be corrected manually in these two files (that's how I did it).

It's manual.

tigrannajaryan (approved):

PR #197, opened 2024-07-09
Allow setting Headers in DownloadableFile message

### LINKED ISSUE #194: Allow setting authentication in `DownloadableFile` message
This issue is spurred by a discussion in https://github.com/open-telemetry/opamp-go/issues/196.

In some cases, it might be useful to have the server include authentication parameters along the `DownloadableFile` in one way or the other.

#### Proposal
The proposed way is to add a new `headers Headers` field, so that can happen via Authentication headers and provide some flexibility into what authentication method is actually used.

#### Alternatives considered
One alternative considered would be to accept a [userinfo](https://datatracker.ietf.org/doc/html/rfc3986#section-3.2.1) section in the `download_url` (like `http://username:password@example.com/`), but that is only good for Basic Auth so it locks us out of using alternative authentication methods, while some newer [HTTP Semantic RFC9110](https://www.rfc-editor.org/rfc/rfc9110#name-deprecation-of-userinfo-in-) advocate for its deprecation so if might not be the best long-term solution

tpaschalis:
Fixes #194 

This PR amends the spec and protos to include an optional set of headers in DownloadableFile, to be used in the download GET request to set up auth parameters. 

This is only one of the ways we could implement setting authentication paramaters to be used by Agents when downloading packages, but it is consistent with other uses of the headers field and generic enough to cover different use cases so I feel it's worth considering.

### DIFF
--- a/proto/opamp.proto
+++ b/proto/opamp.proto
@@ -502,6 +502,13 @@ message DownloadableFile {
     // https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#code-signing
     // for recommendations.
     bytes signature = 3;
+
+    // Optional headers to use when downloading a file. Typically used to set
+    // access tokens or other authorization headers. For HTTP-based protocols
+    // the Agent should set these in the request headers.
+    // For example:
+    // key="Authorization", Value="Basic YWxhZGRpbjpvcGVuc2VzYW1l".
+    Headers headers = 4;
 }
 
 message ServerErrorResponse {
--- a/proto/opamp.proto
+++ b/proto/opamp.proto
@@ -508,6 +508,7 @@ message DownloadableFile {
     // the Agent should set these in the request headers.
     // For example:
     // key="Authorization", Value="Basic YWxhZGRpbjpvcGVuc2VzYW1l".
+    // Status: [Development]
     Headers headers = 4;
 }
 
--- a/specification.md
+++ b/specification.md
@@ -153,6 +153,7 @@ Status: [Beta]
       - [DownloadableFile.download_url](#downloadablefiledownload_url)
       - [DownloadableFile.content_hash](#downloadablefilecontent_hash)
       - [DownloadableFile.signature](#downloadablefilesignature)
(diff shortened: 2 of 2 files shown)


tigrannajaryan (approved):
Thank you @tpaschalis 

LGTM. Once merged let's add corresponding implementation in OpAMP Go.

@andykellr I will keep this open for a couple days in case you have any comments.

tpaschalis:
> Once merged let's add corresponding implementation in OpAMP Go.

Yes, I can handle that.

proto/opamp.proto
```diff
@@ -502,6 +502,13 @@ message DownloadableFile {
     // https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#code-signing
     // for recommendations.
     bytes signature = 3;
+
+    // Optional headers to use when downloading a file. Typically used to set
+    // access tokens or other authorization headers. For HTTP-based protocols
+    // the Agent should set these in the request headers.
+    // For example:
+    // key="Authorization", Value="Basic YWxhZGRpbjpvcGVuc2VzYW1l".
+    Headers headers = 4;
```
tigrannajaryan, line 511:
```suggestion
    // Status: [Development]
    Headers headers = 4;
```

specification.md
```diff
@@ -2343,7 +2344,9 @@ should download the file:
   and the file SHOULD be downloaded from the location specified in the
   [download_url](#downloadablefiledownload_url) field of the
   [DownloadableFile](#downloadablefile-message) message. The Agent SHOULD use an
-  HTTP GET message to download the file.
+  HTTP GET message to download the file. The Agent SHOULD include the HTTP
```
tigrannajaryan, line 2347: Let's mark this addition as `[Development]`. See how it's done in other places.

tigrannajaryan (commented):
One thing I missed is that we want all new additions to start in [Development] status.

specification.md
tpaschalis: Apologies for missing that 🤦

Done, both in the proto definition, and in specification (both inline to the message and in the separate section).

Issue #198, opened 2024-07-09
Agents should report supported components


andykellr:
_I'll describe this in the context of OpenTelemetry, but similar issues will exist for other agents._

OpenTelemetry agents built using the OCB may have different components included during the build. Before sending a configuration to an agent, the management server will want to know what components are available. Sending a configuration with unsupported components will cause an error. It may be desirable to avoid errors and only allow configurations to be applied to agents that are able to use them.

This is similar to agent capabilities but will need more than a bitmask to support arbitrary component names and versions.

This is similar to what is available in component health, but components not in use (but still supported) would not be reported.

This is similar to package statuses but components are no necessarily packages.

This could be reported with remote configuration but doesn't feel right.

andykellr:
Related issue in the collector: https://github.com/open-telemetry/opentelemetry-collector/issues/10570

andykellr:
We discussed in the SIG meeting that a generic `map<string, string> metadata` could be added to the `AgentDescription` message and the structure documented for OpenTelemetry-based agents.

We also discussed making a separate top level `AgentMetadata` message with a separate capability `ReportsMetadata`.

BinaryFissionGames:
I think logically, you'd consider this part of the `AgentDescription`. I'd personally advocate for adding it there if possible.

I think the primary advantage we discussed for the separate message was that it did not need to be sent if the OpAMP server does not use it. I'm not sure if that's a huge advantage or not though, given the `AgentDescription` is generally static and status compression allows it to be omitted most of the time.

BinaryFissionGames:
Regarding the structure of the field, how would you propose representing the components for the collector in a map[string]string? 

Additionally, I feel like arbitrary metadata describing agent is already covered in the identifying/non-identifying attributes, how does this differ?

BinaryFissionGames:
We discussed this again today at the SIG.

We primarily discussed adding this information to the ComponentHealth message, or to the AgentDescription.

BinaryFissionGames:
Here's what I'm thinking in terms of concrete details:

We could imagine doing something like this:
```protobuf
message AgentDescription {
	// ... snip ... //

	// The list of components that are available for this agent to configure. 
	repeated AvailableComponent available_components = 4
}

message AvailableComponent {
	// The ID of the component. This ID MUST be unique within the list of
	// available components.
	string id = 1
	
	// Extra key/value pairs that may be used to describe the component.
	// The key/value pairs are according to semantic conventions, see:
[...]

PR #201, opened 2024-08-26
Add message about available components to AgentDetails

BinaryFissionGames:
This PR adds a new ComponentDetails type that allows agents to communicate metadata relating to the components available in the agent.

### DIFF
--- a/proto/opamp.proto
+++ b/proto/opamp.proto
@@ -708,6 +708,9 @@ enum AgentCapabilities {
     // know the configured interval and should not make assumptions about it.
     // Status: [Development]
     AgentCapabilities_ReportsHeartbeat               = 0x00002000;
+    // The will report AvailableComponents via the AgentToServer.available_components field.
+    // Status: [Development]    
+    AgentCapabilities_ReportsAvailableComponents     = 0x00004000;
     // Add new capabilities here, continuing with the least significant unused bit.
 }
 
--- a/proto/opamp.proto
+++ b/proto/opamp.proto
@@ -168,6 +168,14 @@ message ComponentDetails {
     // Extra key/value pairs that may be used to describe the component.
     // The key/value pairs are according to semantic conventions, see:
     // https://opentelemetry.io/docs/specs/semconv/
+    //
+    // For example, you may use the "code" semantic conventions to
+    // report the location of the code for a specific component:
+    // https://opentelemetry.io/docs/specs/semconv/attributes-registry/code/
+    // 
+    // Or you may use the "vcs" semantic conventions to report the
+    // repository the component may be a part of:
+    // https://opentelemetry.io/docs/specs/semconv/attributes-registry/vcs/
     repeated KeyValue metadata = 1;
 
     // A map of component ID to sub components details. It can nest as deeply as needed to
--- a/proto/opamp.proto
+++ b/proto/opamp.proto
@@ -609,8 +609,22 @@ message AgentDescription {
     //   with this Agent.
     repeated KeyValue non_identifying_attributes = 2;
 
+    // Details about the components available in the agent.
+    ComponentDetails component_details = 3;
+
     // TODO: add ability to specify related entities (such as the Service the Agent is
     // is responsible/associated with).
+}
+
+message ComponentDetails {
+    // Extra key/value pairs that may be used to describe the component.
+    // The key/value pairs are according to semantic conventions, see:
+    // https://opentelemetry.io/docs/specs/semconv/
+    repeated KeyValue metadata = 1;
+
+    // A map of component ID to sub components details. It can nest as deeply as needed to
+    // describe the underlying system.
+    map<string, ComponentDetails> sub_component_map = 2;
 }
 
 enum AgentCapabilities {
--- a/proto/opamp.proto
+++ b/proto/opamp.proto
@@ -610,7 +610,7 @@ message AgentDescription {
     repeated KeyValue non_identifying_attributes = 2;
 
     // Details about the components available in the agent.
-    map<string, ComponentDetails> component_details = 3;
+    map<string, ComponentDetails> available_components = 3;
 
     // TODO: add ability to specify related entities (such as the Service the Agent is
     // is responsible/associated with).
--- a/proto/opamp.proto
+++ b/proto/opamp.proto
@@ -149,7 +149,7 @@ message CertificateRequest {
     bytes csr = 1;
 }
 
-// AvailableComponents contains metadata relating to the components included 
+// AvailableComponents contains metadata relating to the components included
 // within the agent.
 // status: [Development]
 message AvailableComponents {
@@ -172,7 +172,7 @@ message ComponentDetails {
     // For example, you may use the "code" semantic conventions to
     // report the location of the code for a specific component:
     // https://opentelemetry.io/docs/specs/semconv/attributes-registry/code/
-    // 
+    //
     // Or you may use the "vcs" semantic conventions to report the
     // repository the component may be a part of:
     // https://opentelemetry.io/docs/specs/semconv/attributes-registry/vcs/
@@ -256,6 +256,7 @@ enum ServerToAgentFlags {
     // not include the full AvailableComponents message, but only the hash.
     // If this flag is specified, the agent will populate available_components.components
     // with a full description of the agent's components.
+    // Status: [Development]
     ServerToAgentFlags_ReportAvailableComponents = 0x00000002;
 }
 
@@ -709,7 +710,7 @@ enum AgentCapabilities {
     // Status: [Development]
     AgentCapabilities_ReportsHeartbeat               = 0x00002000;
     // The agent will report AvailableComponents via the AgentToServer.available_components field.
-    // Status: [Development]    
+    // Status: [Development]
     AgentCapabilities_ReportsAvailableComponents     = 0x00004000;
     // Add new capabilities here, continuing with the least significant unused bit.
 }
--- a/proto/opamp.proto
+++ b/proto/opamp.proto
@@ -610,7 +610,7 @@ message AgentDescription {
     repeated KeyValue non_identifying_attributes = 2;
 
     // Details about the components available in the agent.
-    ComponentDetails component_details = 3;
+    map<string, ComponentDetails> component_details = 3;
 
     // TODO: add ability to specify related entities (such as the Service the Agent is
     // is responsible/associated with).
--- a/proto/opamp.proto
+++ b/proto/opamp.proto
@@ -708,7 +708,7 @@ enum AgentCapabilities {
     // know the configured interval and should not make assumptions about it.
     // Status: [Development]
     AgentCapabilities_ReportsHeartbeat               = 0x00002000;
-    // The will report AvailableComponents via the AgentToServer.available_components field.
+    // The agent will report AvailableComponents via the AgentToServer.available_components field.
     // Status: [Development]    
     AgentCapabilities_ReportsAvailableComponents     = 0x00004000;
     // Add new capabilities here, continuing with the least significant unused bit.
--- a/proto/opamp.proto
+++ b/proto/opamp.proto
@@ -91,6 +91,10 @@ message AgentToServer {
     // A custom message sent from an Agent to the Server.
     // Status: [Development]
     CustomMessage custom_message = 13;
+
+    // A message indicating the components that are available for configuration on the agent.
+    // Status: [Development]
+    AvailableComponents available_components = 14;
 }
 
 enum AgentToServerFlags {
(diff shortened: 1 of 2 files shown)


proto/opamp.proto
```diff
@@ -609,10 +609,24 @@ message AgentDescription {
     //   with this Agent.
     repeated KeyValue non_identifying_attributes = 2;
 
+    // Details about the components available in the agent.
+    ComponentDetails component_details = 3;
```
BinaryFissionGames, line 613: This mirrors ComponentHealth (single root component), but I'm not sure if there's value in having a "root" ComponentDetails. The root component would just be the agent itself, and I'd expect metadata about the root component to be in identifying/non-identifying attributes.

specification.md
```diff
@@ -1152,6 +1159,124 @@ The following attributes SHOULD be included:
...
+Details about the available components in the agent.
+
+This field gives a description of what components are available, as well
+as extra metadata about the components.
+
+The structure of ComponentDetails DOES NOT need to be a 1-to-1 match with
+the ComponentHealth structure. ComponentHealth generally refers to instances of
+components, while ComponentDetails refers to the available types of components.
+
+This field is not required to be specified.
```
BinaryFissionGames, line 1173: Not sure if this is the right way to word this, but this field should be optional (an agent does not need to specify component_details).

specification.md
```diff
@@ -1152,6 +1159,124 @@ The following attributes SHOULD be included:
...
+```json
+{
+  "sub_component_data": {
+    "receivers": {
+      "sub_component_data": {
+        "github.com/open-telemetry/opentelemetry-collector-contrib/receiver/hostmetricsreceiver@v0.107.0": {
+          "metadata": {
+            "type": "hostmetrics",
+          }
+        }
```
BinaryFissionGames, line 1216: The ID here works for how everything is structured in contrib. I'm wondering if you could cause a collision through, simply by having multiple components in one module? Do we need to support that? I don't think the builder supports that case currently, either.

specification.md
```diff
@@ -1152,6 +1159,124 @@ The following attributes SHOULD be included:
...
+Here's an example of how Fluent Bit could report what components it has available.
+
+```json
+{
+  "sub_component_data": {
+    "input": {
+      "sub_component_data": {
+        "tail": {}
+      }
+    },
```
BinaryFissionGames, line 1258: For most other agents, I expect this map to be very sparse and probably not useful. Each version in fluent-bit, for example, comes with fixed components at fixed versions (so here, the "version" of the component is just the version of fluent-bit, so it's not repeated here).

This area would be more useful for agents that vary in components they include, so agents like the opentelemetry agent (with the builder), or for agents that have a plugin system for dynamic components.

BinaryFissionGames:
@andykellr @tigrannajaryan Here's a rough draft of ComponentDetails, when you get a second to take a look.

specification.md
```diff
@@ -1152,6 +1159,124 @@ The following attributes SHOULD be included:
...
 
+##### AgentDescription.component_details
+
+Details about the available components in the agent.
+
+This field gives a description of what components are available, as well
+as extra metadata about the components.
+
+The structure of ComponentDetails DOES NOT need to be a 1-to-1 match with
+the ComponentHealth structure. ComponentHealth generally refers to instances of
```
tigrannajaryan, line 1170:
```suggestion
the ComponentHealth structure. ComponentHealth generally refers to currently running instances of
```

BinaryFissionGames:
@tigrannajaryan friendly bump on this

specification.md
```diff
@@ -1152,6 +1159,124 @@ The following attributes SHOULD be included:
...
+components, while ComponentDetails refers to the available types of components.
```
tigrannajaryan, line 1171:
```suggestion
components, while ComponentDetails refers to the available types of components, which may not be necessarily running currently. It is also possible that the same component type may have more than one running instance.
```

specification.md
tigrannajaryan, line 1173:
```suggestion
This is an optional field.
```

specification.md
```diff
@@ -1152,6 +1159,124 @@ The following attributes SHOULD be included:
...
+
+A map of component ID to sub components details. It can nest as deeply as needed to
+describe the underlying system.
+
+##### Examples
+
+###### OpenTelemetry Collector
+
+Here is an example of how ComponentDetails could hold information regarding the included
+components for a custom build collector:
```
tigrannajaryan, line 1205:
```suggestion
components for a custom build of [OpenTelemetry Collector](https://github.com/open-telemetry/opentelemetry-collector):
```

tigrannajaryan:
> @tigrannajaryan friendly bump on this

Sorry for the delay. Reviewed.

proto/opamp.proto
```diff
@@ -609,10 +609,24 @@ message AgentDescription {
...
+
     // TODO: add ability to specify related entities (such as the Service the Agent is
     // is responsible/associated with).
 }
 
+message ComponentDetails {
+    // Extra key/value pairs that may be used to describe the component.
+    // The key/value pairs are according to semantic conventions, see:
+    // https://opentelemetry.io/docs/specs/semconv/
```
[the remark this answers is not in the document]
BinaryFissionGames: Yeah, this is a good point. Just thinking about some information you might want:
* Code module
* Code version
* Code repo

There are some code attributes registered with semconv, but I think only `code.namespace` is really applicable here (https://opentelemetry.io/docs/specs/semconv/attributes-registry/code/).

We also have the VCS semantic conventions, which can give e.g. the full URL of the repo and include the ref from which the component was built or taken from, for instance (https://opentelemetry.io/docs/specs/semconv/attributes-registry/vcs/)

I definitely think these should follow the registered semconv where applicable, but it seems like occasionally a component might have some metadata that's more narrow/specific to the agent in a way that might not have a specific semantic convention. For instance, I don't think there's a semantic convention for the "type string" of a component in the opentelemetry collector. In that case, I feel like not using a register semantic convention makes sense.

I'll look into adding some of these into examples.

rbg:
I didn't see when this message would be sent (may be whenever the collector thought it was appropriate (?)); Do we think that possibly the additional feature for a OpAMP server to request this information?

tigrannajaryan:
@BinaryFissionGames please ping me once this is updated to reflect our last discussion in https://github.com/open-telemetry/opamp-spec/issues/198

specification.md
```diff
@@ -2852,6 +2872,163 @@ CustomMessage {
...
+      "null": {},
+      "file": {},
+    }
+  }
+}
+```
+
+#### AvailableComponents.hash
+
+The agent-calculated hash of the components field. The agent MUST include this hash
```
evan-bradley, line 2978: At the end of the day this is an opaque value from the server's point of view, but it would be nice to have some guidance on what this may look like. We should probably at least call out some basics of what properties we expect from the hash ("SHOULD be unique to a particular build of an agent, not to an instance") and that its format is determined by the agent.

specification.md
```diff
@@ -2852,6 +2872,163 @@ CustomMessage {
...
+if it has the ability to report the components it has available.
+
+#### Initial Handshake
```
evan-bradley, line 2981: From my reading of the issue, it sounds like the flow should optionally start with the agent sending the server an AvailableComponents message with just the hash to save a round-trip.

proto/opamp.proto
```diff
@@ -91,6 +91,10 @@ message AgentToServer {
     // A custom message sent from an Agent to the Server.
     // Status: [Development]
     CustomMessage custom_message = 13;
+
+    // A message indicating the components that are available for configuration on the agent.
+    // Status: [Development]
+    AvailableComponents available_components = 14;
```
evan-bradley, line 97: This should have a corresponding capability, right?

specification.md
BinaryFissionGames: :+1: I meant to imply that here, but I don't think I explicitly stated it. I've put a sentence here that makes it clearer.

proto/opamp.proto
BinaryFissionGames: Yes, it definitely should. Added a capability for ReportsAvailableComponents

specification.md
BinaryFissionGames: I added that the agent hash should be the same if the components are the same. 

I want to avoid saying it should be unique for a build of the agent, since I'd like this to also support plugin-based agents where a particular build might be the same but the plugins on the filesystem might differ.

specification.md
```diff
@@ -2852,6 +2872,163 @@ CustomMessage {
...
+
+In order to reduce the amount of data transmitted, the AvailableComponents message
+does not initially contain the entire components map. Instead, the AvailableComponents
+message will have the agent computed hash set, with an empty map for components.
+
+The OpAMP server may use this hash to determine whether it remembers the set of
+AvailableComponents or not. If the hash is not found on the OpAMP server, the server
+may request the full components map is reported by setting the ReportAvailableComponents
```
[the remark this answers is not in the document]
BinaryFissionGames: Updated to this wording, seems good to me!

proto/opamp.proto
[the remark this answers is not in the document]
BinaryFissionGames: I update to point out these conventions (code, vcs).

specification.md
BinaryFissionGames: This is a good point.

I inverted the ID to be the type string instead; I think the type should be unique within it's class (e.g. there is only one "hostmetricsreceiver" in any given build of the agent). Then the module is in the metadata.

BinaryFissionGames:
Looks like a link is broken unrelated to the changes in this PR, but this should be good for another look.

andykellr (approved):

proto/opamp.proto
```diff
@@ -663,6 +708,9 @@ enum AgentCapabilities {
     // know the configured interval and should not make assumptions about it.
     // Status: [Development]
     AgentCapabilities_ReportsHeartbeat               = 0x00002000;
+    // The will report AvailableComponents via the AgentToServer.available_components field.
```
evan-bradley, line 711:
```suggestion
    // The agent will report AvailableComponents via the AgentToServer.available_components field.
```

specification.md
```diff
@@ -2852,6 +2873,168 @@ CustomMessage {
...
+This field may be omitted from the message if the OpAMP server has not explicitly
+requested it by setting the ReportAvailableComponents flag in the preceding
+ServerToAgent message.
+
+##### Examples
+
+###### OpenTelemetry Collector
+
+Here is an example of how ComponentDetails could hold information regarding the included
+components for a custom build of [OpenTelemetry Collector](https://github.com/open-telemetry/opentelemetry-collector):
```
evan-bradley, line 2906:
```suggestion
components for a custom build of the [OpenTelemetry Collector](https://github.com/open-telemetry/opentelemetry-collector):
```

evan-bradley (approved):

proto/opamp.proto
```diff
@@ -212,6 +251,12 @@ enum ServerToAgentFlags {
     // AgentToServer.sequence_num values.
     // The Server asks the Agent to report full status.
     ServerToAgentFlags_ReportFullState = 0x00000001;
+
+    // ReportAvailableComponents flag can be used by the server if the Agent did
+    // not include the full AvailableComponents message, but only the hash.
+    // If this flag is specified, the agent will populate available_components.components
+    // with a full description of the agent's components.
+    ServerToAgentFlags_ReportAvailableComponents = 0x00000002;
```
tigrannajaryan, line 259: Please mark `Status: [Development]`.

specification.md
```diff
@@ -2852,6 +2873,168 @@ CustomMessage {
...
+
+The server may then optionally persist the components for this hash, so that the
+server does not need to request them again on subsequent connects.
+
+The AvailableComponents message is also subject to [status compression](#agent-status-compression),
+and may be omitted if the hash has not changed from its previous value.
+
+#### ComponentDetails Message
+
+Status: [Beta]
```
tigrannajaryan, line 3007:
```suggestion
Status: [Development]
```

Or perhaps omit it since the parent section is already `[Development]` (I think).

tigrannajaryan:
@andykellr or @dpaasman00 will you take over this PR to make the final fixes before merging?

specification.md
andykellr: > Or perhaps omit it since the parent section is already `[Development]` (I think).

I decided to omit it.

proto/opamp.proto
andykellr: Done

andykellr:
> @andykellr or @dpaasman00 will you take over this PR to make the final fixes before merging?

Yes, I just pushed changes based on your feedback. It should be ready.

PR #205, opened 2024-10-15
Add TLS settings to all connection settings

### LINKED ISSUE #203: Proxy & TLS settings in OpAMPConnectionSettings, TelemetryConnectionSettings, and OtherConnectionSettings
Connection settings that an OpAMP server offers should have some way to pass additional information that includes proxy settings and additional TLS settings.

TLS settings should TLS/mTLS settings defined in [otel config tls](https://github.com/open-telemetry/opentelemetry-collector/tree/main/config/confighttp#http-configuration-settings).

Proxy settings may include:
- proxy url
- proxy headers
- proxy client certificate
- proxy TLS settings

Additionally I think that we should include `other_settings` in the definition of OpAMPConnectionSettings and TelemetryConnectionSettings to allow implementation to define their own additional settings.

michel-laterman:
Add new `TLSConnectionSettings` across all connection settings a server can offer.

- Relates #203

### DIFF
--- a/proto/opamp.proto
+++ b/proto/opamp.proto
@@ -919,7 +919,7 @@ message PackageDownloadDetails {
     double download_percent = 1;
 
     // The current package download rate in bytes per second.
-    uint64 download_bytes_per_second = 2;
+    double download_bytes_per_second = 2;
 }
 
 // The status of this package.
--- a/proto/opamp.proto
+++ b/proto/opamp.proto
@@ -126,6 +126,12 @@ message ConnectionSettingsRequest {
     // for the Server.
     OpAMPConnectionSettingsRequest opamp = 1;
 
+
+    // A request for the server to send a message that includes
+    // ConnectionSettingsOffers.
+    // Status: [Development]
+    SettingsRequest settings_request = 2;
+
     // In the future we can add request fields for non-OpAMP connection types
     // (own telemetry, other connections).
 }
@@ -147,6 +153,17 @@ message CertificateRequest {
     // OpAMPConnectionSettings where the certificate.cert contains the issued
     // certificate.
     bytes csr = 1;
+}
+
+// SettingsRequest is a request for the Server to populate ConnectionSettingsOffers
+// in its response.
+// Status: [Development]
+message SettingsRequest {
+  // The latest ConnectionSettingsOffers.hash the client has.
+  // An empty hash indicates that the client has not recieved any connection settings.
+  // The server SHOULD compare the hash to the currently offered settings hash and
+  // include the new settings if the hash has changed.
+  bytes hash = 1;
 }
 
 // AvailableComponents contains metadata relating to the components included
--- a/proto/opamp.proto
+++ b/proto/opamp.proto
@@ -91,6 +91,10 @@ message AgentToServer {
     // A custom message sent from an Agent to the Server.
     // Status: [Development]
     CustomMessage custom_message = 13;
+
+    // A message indicating the components that are available for configuration on the agent.
+    // Status: [Development]
+    AvailableComponents available_components = 14;
 }
 
 enum AgentToServerFlags {
@@ -144,6 +148,41 @@ message CertificateRequest {
     // certificate.
     bytes csr = 1;
 }
+
+// AvailableComponents contains metadata relating to the components included
+// within the agent.
+// status: [Development]
+message AvailableComponents {
+    // A map of a unique component ID to details about the component.
+    // This may be omitted from the message if the server has not
+    // explicitly requested it be sent by setting the ReportAvailableComponents
+    // flag in the previous ServerToAgent message.
+    map<string, ComponentDetails> components = 1;
+
+    // Agent-calculated hash of the components.
+    // This hash should be included in every AvailableComponents message.
+    bytes hash = 2;
+ }
+
+message ComponentDetails {
+    // Extra key/value pairs that may be used to describe the component.
+    // The key/value pairs are according to semantic conventions, see:
+    // https://opentelemetry.io/docs/specs/semconv/
+    //
+    // For example, you may use the "code" semantic conventions to
+    // report the location of the code for a specific component:
+    // https://opentelemetry.io/docs/specs/semconv/attributes-registry/code/
+    //
+    // Or you may use the "vcs" semantic conventions to report the
+    // repository the component may be a part of:
+    // https://opentelemetry.io/docs/specs/semconv/attributes-registry/vcs/
+    repeated KeyValue metadata = 1;
+
+    // A map of component ID to sub components details. It can nest as deeply as needed to
+    // describe the underlying system.
+    map<string, ComponentDetails> sub_component_map = 2;
+}
+
 
 message ServerToAgent {
     // Agent instance uid. MUST match the instance_uid field in AgentToServer message.
@@ -212,6 +251,13 @@ enum ServerToAgentFlags {
     // AgentToServer.sequence_num values.
     // The Server asks the Agent to report full status.
     ServerToAgentFlags_ReportFullState = 0x00000001;
+
+    // ReportAvailableComponents flag can be used by the server if the Agent did
+    // not include the full AvailableComponents message, but only the hash.
+    // If this flag is specified, the agent will populate available_components.components
+    // with a full description of the agent's components.
+    // Status: [Development]
+    ServerToAgentFlags_ReportAvailableComponents = 0x00000002;
 }
 
 enum ServerCapabilities {
@@ -699,6 +745,9 @@ enum AgentCapabilities {
     // know the configured interval and should not make assumptions about it.
     // Status: [Development]
     AgentCapabilities_ReportsHeartbeat               = 0x00002000;
+    // The agent will report AvailableComponents via the AgentToServer.available_components field.
+    // Status: [Development]
+    AgentCapabilities_ReportsAvailableComponents     = 0x00004000;
     // Add new capabilities here, continuing with the least significant unused bit.
 }
 
--- a/proto/opamp.proto
+++ b/proto/opamp.proto
@@ -277,6 +277,16 @@ message OpAMPConnectionSettings {
     // If this field has no value or is set to 0, the Agent should not send any heartbeats.
     // Status: [Development]
     uint64 heartbeat_interval_seconds = 4;
+
+    // Additional connection settings. These are Agent-specific and are up to the Agent
+    // interpret.
+    map<string, string> other_settings = 5;
+
+    // Optional connection specific TLS settings.
+    TLSConnectionSettings tls = 6;
+
+    // Optional connection specific proxy settings.
+    ProxyConnectionSettings proxy = 7;
 }
 
 // The TelemetryConnectionSettings message is a collection of fields which comprise an
@@ -303,6 +313,16 @@ message TelemetryConnectionSettings {
     // This field is optional: if omitted the client SHOULD NOT use a client-side certificate.
     // This field can be used to perform a client certificate revocation/rotation.
     TLSCertificate certificate = 3;
+
+    // Additional connection settings. These are Agent-specific and are up to the Agent
+    // interpret.
+    map<string, string> other_settings = 4;
+
+    // Optional connection specific TLS settings.
+    TLSConnectionSettings tls = 5;
+
+    // Optional connection specific proxy settings.
+    ProxyConnectionSettings proxy = 6;
 }
 
 // The OtherConnectionSettings message is a collection of fields which comprise an
@@ -348,6 +368,71 @@ message OtherConnectionSettings {
     // Other connection settings. These are Agent-specific and are up to the Agent
     // interpret.
     map<string, string> other_settings = 4;
+
+    // Optional connection specific TLS settings.
+    TLSConnectionSettings tls = 5;
+
+    // Optional connection specific proxy settings.
+    ProxyConnectionSettings proxy = 6;
+}
+
+// Status: [Beta]
+message TLSConnectionSettings {
+  // Insecure is false by default, if true TLS will be disabled for the connection.
+  bool insecure = 1;
+
+  // Optional path to the CA file on disk
+  string ca_file = 2;
+
+  // Alternative to ca_file, Provides CA cert contents as a string
+  string ca_pem = 3;
+
+  // Load system CA pool alongside any specifed CAs (provided through ca_file or ca_path).
+  bool include_system_ca_certs_pool = 4;
+
+  // skip certificate verification
+  bool insecure_skip_verify = 5;
+
+  // Miniumum accepted TLS version; default "1.2".
+  string min_version = 6;
+
+  // Maxiumum accepted TLS version; default "".
+  string max_version = 7;
+
+  // Explicit list of cipher suites.
+  repeated string cipher_suites = 8;
+}
+
+// Status: [Beta]
+message ProxyConnectionSettings {
+    // disable using a proxy for the connection.
+    // If true configured settings, and environment variables are ignored for the connection.
+    bool disable = 1;
+
+    // A URL, host:port or some other destination specifier.
+    string destination_endpoint = 2;
+
+    // Optional headers to use when connecting. Typically used to set access tokens or
+    // other authorization headers. For HTTP-based protocols the Agent should
+    // set these in the request headers.
+    // For example:
+    // key="Authorization", Value="Basic YWxhZGRpbjpvcGVuc2VzYW1l".
+    Headers headers = 3;
+
+    // The Agent should use the offered certificate to connect to the destination
+    // from now on. If the Agent is able to validate and connect using the offered
+    // certificate the Agent SHOULD forget any previous client certificates
+    // for this connection.
+    // This field is optional: if omitted the client SHOULD NOT use a client-side certificate.
+    // This field can be used to perform a client certificate revocation/rotation.
+    TLSCertificate certificate = 4;
+
+    // Other connection settings. These are Agent-specific and are up to the Agent
+    // interpret.
+    map<string, string> other_settings = 5;
+
+    // Connection specific TLS settings.
+    TLSConnectionSettings tls = 6;
 }
 
 // Status: [Beta]
--- a/proto/opamp.proto
+++ b/proto/opamp.proto
@@ -427,26 +427,23 @@ message OtherConnectionSettings {
 
 // Status: [Development]
 message TLSConnectionSettings {
-  // Insecure is false by default, if true TLS will be disabled for the connection.
-  bool insecure = 1;
+  // Provides CA cert contents as a string.
+  string ca_pem_contents = 1;
 
-  // Alternative to ca_file_path, Provides CA cert contents as a string
-  string ca_pem_contents = 2;
+  // Load system CA pool alongside any passed CAs.
+  bool include_system_ca_certs_pool = 2;
 
-  // Load system CA pool alongside any specifed CAs (provided through ca_file or ca_path).
-  bool include_system_ca_certs_pool = 3;
-
-  // skip certificate verification
-  bool insecure_skip_verify = 4;
+  // skip certificate verification.
+  bool insecure_skip_verify = 3;
 
   // Miniumum accepted TLS version; default "1.2".
-  string min_version = 5;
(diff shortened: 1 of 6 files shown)


proto/opamp.proto
```diff
@@ -277,6 +277,16 @@ message OpAMPConnectionSettings {
     // If this field has no value or is set to 0, the Agent should not send any heartbeats.
     // Status: [Development]
     uint64 heartbeat_interval_seconds = 4;
+
+    // Additional connection settings. These are Agent-specific and are up to the Agent
```
tigrannajaryan, line 281: Mark all new fields and messages with `Status: [Development]`. New features start at this level.

tigrannajaryan: I [posted](https://github.com/open-telemetry/opamp-spec/issues/203#issuecomment-2399956499) earlier:
>One other design consideration is to make sure OpAMPConnectionSettings, TelemetryConnectionSettings and OtherConnectionSettings are defined uniformly, with the exception of settings which are only applicable to a particular connection type (e.g. heartbeat_interval_seconds is unique to OpAMPConnectionSettings). A possible option would be to move all common settings to one proto message and include that from OpAMPConnectionSettings, TelemetryConnectionSettings and OtherConnectionSettings. That would be a breaking change, so we will need to discuss if it is worth it.

@andykellr what do you think?

proto/opamp.proto
```diff
@@ -348,6 +368,71 @@ message OtherConnectionSettings {
...
+    // This field is optional: if omitted the client SHOULD NOT use a client-side certificate.
+    // This field can be used to perform a client certificate revocation/rotation.
+    TLSCertificate certificate = 4;
+
+    // Other connection settings. These are Agent-specific and are up to the Agent
+    // interpret.
+    map<string, string> other_settings = 5;
+
+    // Connection specific TLS settings.
+    TLSConnectionSettings tls = 6;
```
tigrannajaryan, line 435: Why TLSConnectionSettings here? We already have one as a sibling of ProxyConnectionSettings.

proto/opamp.proto
```diff
@@ -348,6 +368,71 @@ message OtherConnectionSettings {
...
+
+    // A URL, host:port or some other destination specifier.
+    string destination_endpoint = 2;
+
+    // Optional headers to use when connecting. Typically used to set access tokens or
+    // other authorization headers. For HTTP-based protocols the Agent should
+    // set these in the request headers.
+    // For example:
+    // key="Authorization", Value="Basic YWxhZGRpbjpvcGVuc2VzYW1l".
+    Headers headers = 3;
```
tigrannajaryan, line 420: Are these headers in addition to headers in OpAMPConnectionSettings? Do we need both?

proto/opamp.proto
```diff
@@ -348,6 +368,71 @@ message OtherConnectionSettings {
...
+
+    // The Agent should use the offered certificate to connect to the destination
+    // from now on. If the Agent is able to validate and connect using the offered
+    // certificate the Agent SHOULD forget any previous client certificates
+    // for this connection.
+    // This field is optional: if omitted the client SHOULD NOT use a client-side certificate.
+    // This field can be used to perform a client certificate revocation/rotation.
+    TLSCertificate certificate = 4;
```
tigrannajaryan, line 428: Why TLSCertificate here? We already have one as a sibling of ProxyConnectionSettings.

tigrannajaryan (commented):
Thanks for the draft @michel-laterman 

I left some comments, and I would like @andykellr to also review.

proto/opamp.proto
michel-laterman: I think it would help make it clear for cases where a proxy requires these settings  but the actual endpoint being reached does not

proto/opamp.proto
andykellr: I would definitely like to avoid breaking changes at this point. I think the approach of having additional messages for TLS and Proxy settings creates enough uniformity between the connection settings messages.

I likely won't be able to review this PR until early next week, but I will take a look. At first glance it looks reasonable but you raise good questions.

tigrannajaryan: OK, let's avoid the breaking changes.

proto/opamp.proto
```diff
@@ -303,6 +315,18 @@ message TelemetryConnectionSettings {
     // This field is optional: if omitted the client SHOULD NOT use a client-side certificate.
     // This field can be used to perform a client certificate revocation/rotation.
     TLSCertificate certificate = 3;
+
+    // Additional connection settings. These are Agent-specific and are up to the Agent
+    // interpret.
+    map<string, string> other_settings = 4;
```
tigrannajaryan, line 321: See my comment on the [other PR](https://github.com/open-telemetry/opamp-spec/pull/206#discussion_r1811346270). I would like to understand what agent-specific data we intend to record here.

tigrannajaryan (changes requested):
I would advise to break this down into 2 PRs: one that adds `TLSConnectionSettings`, the other that adds `ProxyConnectionSettings`.

If there are known uses cases for `other_settings` that should likely be the 3rd PR where we can discuss it.

proto/opamp.proto
michel-laterman: Adding `other_settings` was to make the connection types more consistent. I'll limit the scope of my PRs

michel-laterman:
I'll recreate the ProxySettings in another pr

proto/opamp.proto
```diff
@@ -348,6 +356,37 @@ message OtherConnectionSettings {
...
+// Status: [Development]
+message TLSConnectionSettings {
+  // Insecure is false by default, if true TLS will be disabled for the connection.
+  bool insecure = 1;
+
+  // Optional path to the CA file on disk
+  string ca_file = 2;
+
+  // Alternative to ca_file, Provides CA cert contents as a string
+  string ca_pem = 3;
```
tigrannajaryan, line 374: How do we expect this to work? Is OpAMP servers expected to know where on agent's disk these files are located? How would the server know?

This also potentially is an attack vector where malicious server can direct the agent to read an arbitrary file. On its own it may seem benign but if we end up having another weakness where the file that was read for some reasons gets reported to the server this will allow the server to remotely read arbitrary files from agent's machine. This goes against our zero trust [recommendations](https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#general-recommendations) in the Security section.

Should we instead send the _content_ of CA certificate to the agent to use? The agent can save the content locally and use it in subsequent connections.

This can work particularly well for `TelemetryConnectionSettings` and `OtherConnectionSettings`, the idea being that OpAMP server has the knowledge about the CA used by telemetry and other connection destinations.

And even for `OpAMPConnectionSettings` this may make sense as a flow, where the agent first connects without the TLS certificate, in sort of a bootstraping mode, then accepts server offered CA certificate for future verifications. This is similar to Trust On First Use flow that we already have but in the opposite direction (trust of server, instead of trust of client).

If we think this is a valid usage of TLSConnectionSettings I think we should add a corresponding section in the spec and include a sequence diagram that explains it.

proto/opamp.proto
```diff
@@ -348,6 +356,37 @@ message OtherConnectionSettings {
...
+    TLSConnectionSettings tls = 5;
+}
+
+// Status: [Development]
+message TLSConnectionSettings {
+  // Insecure is false by default, if true TLS will be disabled for the connection.
+  bool insecure = 1;
+
+  // Optional path to the CA file on disk
+  string ca_file = 2;
```
tigrannajaryan, line 371:
```suggestion
  string ca_file_path = 2;
```

proto/opamp.proto
michel-laterman: When specifying the file path we assume it is present at that location across all agents. It can be provisioned by some other tool (baked into the image, or by a configuration management tool).

But I can see how having a filepath specification goes against our current recommendations. It's specified because it is a common approach to TLS configuration, however I can remove the it if we want to keep our security exposure low.

The content can instead be specified with the `ca_pem` attribute (I'll rename to `ca_pem_content` to make that clearer).

I like the idea of a TOFU like mechanism for accepting CAs from servers, I'll try to add the description for this in the spec.
At elastic we do something a little similar to that, we allow the specification of a CA by a fingerprint hash and add the matching CA to the cert pool when it is found in a connection ([implementation here](https://github.com/elastic/elastic-agent-libs/blob/main/transport/tlscommon/tls_config.go#L177)). This assumes that what ever is installing an agent has some additional information about the server and it prevents an agent from blindly trusting any CA. This may be worth breaking out into another issue/pr, what do you think?

tigrannajaryan: >however I can remove the it if we want to keep our security exposure low.

Yes, I think we should remove it for now.

proto/opamp.proto
```diff
@@ -348,6 +356,37 @@ message OtherConnectionSettings {
...
+
+    // Optional connection specific TLS settings.
+    // Status: [Development]
+    TLSConnectionSettings tls = 5;
+}
+
+// Status: [Development]
+message TLSConnectionSettings {
+  // Insecure is false by default, if true TLS will be disabled for the connection.
+  bool insecure = 1;
```
tigrannajaryan, line 368: What is the use case for this flag? Why would you want to remotely specify TLS connection settings and at the same time tell the agent to ignore them?

Generally speaking the use cases are not entirely clear. How are all these parameters supposed to be used? It seems like we borrowed the settings from the Collector which is probably a good starting point but we need to understand if they are applicable to OpAMP-managed configuration.

If we don't have a clear use case for a setting then we should not include it here.

michel-laterman: The best use case I can think of for this case is to ease development/testing (allow http connections)

tigrannajaryan:
@michel-laterman we had a discussion with other OpAMP approvers/maintainers and decided to follow the Otel spec's requirements for making new proposals, namely to ask for prototypes that demonstrate the new capabilities. We will be formalizing the requirements in this PR: https://github.com/open-telemetry/opamp-spec/pull/207

I think it is important for this particular proposal to show how the TLS settings will be used. It is not a trivial change, so a working code would help understand it better.

michel-laterman:
@tigrannajaryan, I have a WIP demo for offering the TLS settings: https://github.com/open-telemetry/opamp-go/pull/338

I have yet to add a demo for Agent-initiated CA trust Flow.

specification.md
```diff
@@ -1579,6 +1584,57 @@ immediately after successful connection each Agent will acquire their own unique
 connection credentials. This way individual Agent's credentials may be revoked
 without disrupting the access to all other Agents.
 
+#### Agent-initiated CA trust Flow
+
+Status: [Development]
+
+This is an Agent-initiated flow that allows an agent to connect to a Server without
```
michel-laterman, line 1591: Right now this does not use an explicit signal (from the agent to the server).

We can add one as a part of the `OpAMPConnectionSettingsRequest` message.
But then I'm concerned that we don't have a way for the agent to signal to the server that offered settings have been attempted, but have failed.

michel-laterman:
To give a rough recap on my question about initial CA distribution during today's opamp sig meeting.

The current spec as written uses an implicit workflow where the client will connect with `InsecureSkipVerify: true` and the server will send offer connection settings to a new agent in response to the initial connection.
My demo implementation: https://github.com/open-telemetry/opamp-go/pull/338 does this if both the client and server are started with special flags (`-await-ca/-send-ca`)

The idea I have for changing this is to make it explicit behaviour. We add a new attribute to `OpAMPConnectionSettingsRequest`  (called something like `ConnectionSettingsRequest`) that a client can use to demand `ConnectionSettingsOffers` from the server.
With this approach, in order to trigger the setting being sent, the client would be started with a new flag (i.e., `-request-connection-settings`), and would connect to the opamp server with `InsecureSkipVerify: true` initially, but include `OpAMPConnectionSettingsRequest.ConnectionSettingsRequest` in the first message sent to the server

michel-laterman:
I went ahead and made the intial CA distribution require a explicit signal from a client.
I have already added to my demo: https://github.com/open-telemetry/opamp-go/pull/338
I'm going to mark this as ready for review

specification.md
```diff
@@ -1634,6 +1640,69 @@ immediately after successful connection each Agent will acquire their own unique
...
+#### Agent-initiated CA trust Flow
+
+Status: [Development]
+
+This is an Agent-initiated flow that allows an agent to connect to a Server
+without trusting its CA. Initially the agent will connect without any form of
+verification, the initial AgentToServer message will indicate that the agent
+expects connection connection settings to be offered in the response message.
+The response contains the PEM content associated with its CA certificate as part
+of the ServerToAgent response using the connection settings offering.
```
andykellr, line 1652:
```suggestion
This is an Agent-initiated flow that allows an agent to connect to a Server
without trusting its CA. Initially the agent will connect without any form of
verification. The initial AgentToServer message will indicate using SettingsRequest that the agent
expects connection settings to be offered in the response message.
The response contains the PEM content associated with its CA certificate as part
of the ServerToAgent response using the connection settings offering.
```

andykellr:
Can you describe the use of the `hash`? It isn't used in the example implementation. When would the client request settings again? Also is the hash stored somewhere on the client?

The example implementation in opamp-go uses a sync.Once called `requestConnectionSettingsOnce` to ensure that there is only one request for the connection settings. If we only expect this to happen once then it should probably be stated in the spec. Regarding that implementation, I think there could be a case where the connection is dropped or the server fails to respond (could send 503 Service Unavailable) and the client will not attempt to request the settings again because the Once will have completed. I think the spec should be more clear about the circumstances under which the client will/should send a SettingsRequest.

proto/opamp.proto
```diff
@@ -394,6 +419,31 @@ message OtherConnectionSettings {
...
+// Status: [Development]
+message TLSConnectionSettings {
+  // Provides CA cert contents as a string.
+  string ca_pem_contents = 1;
+
+  // Load system CA pool alongside any passed CAs.
+  bool include_system_ca_certs_pool = 2;
+
+  // skip certificate verification.
+  bool insecure_skip_verify = 3;
```
tigrannajaryan, line 437: Isn't this a must if the server sends a new CA certificate? The agent has no way to verify it anway since the CA is unknown to the agent.

proto/opamp.proto
```diff
@@ -394,6 +419,31 @@ message OtherConnectionSettings {
...
+    TLSConnectionSettings tls = 5;
+}
+
+// Status: [Development]
+message TLSConnectionSettings {
+  // Provides CA cert contents as a string.
+  string ca_pem_contents = 1;
+
+  // Load system CA pool alongside any passed CAs.
+  bool include_system_ca_certs_pool = 2;
```
tigrannajaryan, line 434: This doesn't seem to be needed for the new flow.

proto/opamp.proto
```diff
@@ -394,6 +419,31 @@ message OtherConnectionSettings {
...
+  bool insecure_skip_verify = 3;
+
+  // Miniumum accepted TLS version; default "1.2".
+  string min_version = 4;
+
+  // Maxiumum accepted TLS version; default "".
+  string max_version = 5;
+
+  // Explicit list of cipher suites.
+  repeated string cipher_suites = 6;
```
tigrannajaryan, line 446: These seem to be unrelated to the new flow.

specification.md
```diff
@@ -1634,6 +1640,69 @@ immediately after successful connection each Agent will acquire their own unique
 connection credentials. This way individual Agent's credentials may be revoked
 without disrupting the access to all other Agents.
 
+#### Agent-initiated CA trust Flow
+
+Status: [Development]
+
+This is an Agent-initiated flow that allows an agent to connect to a Server
+without trusting its CA. Initially the agent will connect without any form of
```
tigrannajaryan, line 1648: It is not really "without trusting". The Agent will trust the CA once the flow is completed. It is just that the Agent does not have the CA certificate initially. Perhaps reword it like this?
```suggestion
without have the possession of the Server's CA initially. Initially the agent will connect without any form of
```

specification.md
```diff
@@ -1634,6 +1640,69 @@ immediately after successful connection each Agent will acquire their own unique
...
+verification, the initial AgentToServer message will indicate that the agent
+expects connection connection settings to be offered in the response message.
```
tigrannajaryan, line 1650:
```suggestion
expects connection settings to be offered in the response message.
```

specification.md
```diff
@@ -1795,12 +1864,15 @@ and respond with an offer of connection settings for the Agent.
...
 message ConnectionSettingsRequest {
     OpAMPConnectionSettingsRequest opamp = 1;
+    SettingsRequest settings_request = 2;
 }
 ```
 
 The `opamp` field is set to indicate a request for OpAMP connection settings.
-If this field is unset then the ConnectionSettingsRequest message is empty and is
-not actionable for the Server.
+The `settings_request` field is set to indicate a request for all offered
```
tigrannajaryan, line 1872: "for all" or specifically "for TLSConnectionSettings"? What does "all" mean in this context?

proto/opamp.proto
michel-laterman: For the flow, no.
Do we want the settings to only support the flow, or to correspond more with general TLS settings?
I think having more generic settings present will help in the future

proto/opamp.proto
michel-laterman: Again, this setting is meant to be more generic and not strictly tied to the CA distribution flow.

specification.md
michel-laterman: All settings for OpAMP  connection (this may include the hosts, and things like the heartbeat)

michel-laterman:
> Can you describe the use of the `hash`? It isn't used in the example implementation. When would the client request settings again? Also is the hash stored somewhere on the client?
> 

@andykellr `hash` for this is meant to match `ConnectionSettingsOffers.hash` there's references to it in opamp-go, but I don't think it's used at the moment.

> The example implementation in opamp-go uses a sync.Once called `requestConnectionSettingsOnce` to ensure that there is only one request for the connection settings. If we only expect this to happen once then it should probably be stated in the spec. Regarding that implementation, I think there could be a case where the connection is dropped or the server fails to respond (could send 503 Service Unavailable) and the client will not attempt to request the settings again because the Once will have completed. I think the spec should be more clear about the circumstances under which the client will/should send a SettingsRequest.

Good point, I'll remove the use of `Once` and leave it purely up to the client

proto/opamp.proto
tigrannajaryan: We need to have specific use cases in mind that require these settings. My preference is to avoid adding settings that don't have clear uses and important cases.

specification.md
```diff
@@ -1634,6 +1640,70 @@ immediately after successful connection each Agent will acquire their own unique
 connection credentials. This way individual Agent's credentials may be revoked
 without disrupting the access to all other Agents.
 
+#### Agent-initiated CA trust Flow
+
+Status: [Development]
+
+This is an Agent-initiated flow that allows an agent to connect to a Server
+without the possesion of the Server's CA initially. Initially the agent will
```
tigrannajaryan, line 1648: BTW, I think this same flow can also work for _replacing_ the CA. So even if the Agent knows a CA that the Server uses initially, the Server can offer a different CA to be used in the future. The difference will be that the initial connection will be with CA verification. Do we want to generalize this flow to also allow this use case?

specification.md
```diff
@@ -2074,9 +2093,35 @@ This field can be used to perform a client certificate revocation/rotation.
...
+Optional connection specific TLS settings.
+
+#### TLSConnectionSettings Message
+
+Status: [Development]
+
+The message carries optional TLS settings that are used to configure a client's
+connection. If the Agent is able to validate the connection settings, the Agent
+SHOULD forget any previous TLS settings. If this message is not included, the
+client SHOULD use the agent's default TLS settings for the connection.
```
michel-laterman, line 2109: @tigrannajaryan @andykellr, I've added a description on how TLS settings should function, and have changed the demo (https://github.com/open-telemetry/opamp-go/pull/338) to reflect this.

The goal is that when `TLSConnectionSettings` is present, it can form the TLS config (excluding client certificate).
If no TLS settings are sent with a connection offering, the agent's defaults should be used.

andykellr (approved):
I'll let @tigrannajaryan merge when he has a chance to review.

tigrannajaryan (approved):

tigrannajaryan:
Thanks for you patience @michel-laterman

PR #206, opened 2024-10-16
Add downloading state and metadata to PackageStatus

### LINKED ISSUE #204: PackageStatus message should define a downloading state and include additional metadata
[PackageStatus messages](https://opentelemetry.io/docs/specs/opamp/#packagestatus-message) should include an optional `DOWNLOADING` status that may be used by a client to signal that it is downloading the package.
Additionally we allow for an optional metadata attribute (`map[string]string`) to allow additional details related to the state to be included, such as a message, download_rate, download_percent, etc.
metadata should also be optional for the existing states (`INSTALLED`, `INSTALLING`, `INSTALL_FAILED`) and may include relevant information.

This will allow users to determine if the install is taking a long time due to a network issue instead of only having the `INSTALLING` state appear.

michel-laterman:
Add a separate `Downloading` `PackageStatusEnum` so an agent can use a distinguish between downloading and installing a new package.
Add a `metadata` attribute that may optionally be used with any `PackageStatus` message to give a user additional details which may include things like a download rate, percentage download, or messages about the current state.

- Closes #204

### DIFF
--- a/proto/opamp.proto
+++ b/proto/opamp.proto
@@ -802,6 +802,10 @@ message PackageStatus {
 
     // Error message if the status is erroneous.
     string error_message = 7;
+
+    // Optional details that may be of interest to a user.
+    // For example the download rate, percent downloaded, or a message.
+    map<string, string> metadata = 8;
 }
 
 // The status of this package.
@@ -814,7 +818,7 @@ enum PackageStatusEnum {
     // Installation of this package has not yet started.
     PackageStatusEnum_InstallPending = 1;
 
-    // Agent is currently downloading and installing the package.
+    // Agent is currently installing the package.
     // server_offered_hash field MUST be set to indicate the version that the
     // Agent is installing. The error_message field MUST NOT be set.
     PackageStatusEnum_Installing = 2;
@@ -824,6 +828,11 @@ enum PackageStatusEnum {
     // tried to install. The error_message may also contain more details about
     // the failure.
     PackageStatusEnum_InstallFailed = 3;
+
+    // Agent is currently downloading the package.
+    // server_offered_hash field MUST be set to indicate the version that the
+    // Agent is installing. The error_message field MUST NOT be set.
+    PackageStatusEnum_Downloading = 4;
 }
 
 // Properties related to identification of the Agent, which can be overridden


proto/opamp.proto
```diff
@@ -802,6 +802,11 @@ message PackageStatus {
 
     // Error message if the status is erroneous.
     string error_message = 7;
+
+    // Optional details that may be of interest to a user.
+    // For example the download rate, percent downloaded, or a message.
+    // Status: [Development]
+    map<string, string> metadata = 8;
```
tigrannajaryan, line 809: I am not sure about this one. Since we don't define what the metadata is going to be it becomes purely agent-specific. When possible we should avoid agent-specific additions since it means servers have to implement agent-specific logic. The dowloading of a package seems to be a common concern, so I don't see at the moment why it needs to be an open map with "anything possible to record" approach.

I would like to see use cases that demonstrate that every agent is going to use different metadata before we can go ahead with this and there is nothing or very little common between agent types.

Absent the use cases I suggest we do one of the following:
1. Delete this from the PR so that we can merge it with the other change, which is not controversial.
1. Replace this by a more concrete set of details that describe the state of downloading (e.g. bytes downloaded, ETA, etc).

proto/opamp.proto
```diff
@@ -824,6 +829,11 @@ enum PackageStatusEnum {
     // tried to install. The error_message may also contain more details about
     // the failure.
     PackageStatusEnum_InstallFailed = 3;
+
+    // Agent is currently downloading the package.
+    // server_offered_hash field MUST be set to indicate the version that the
+    // Agent is installing. The error_message field MUST NOT be set.
+    PackageStatusEnum_Downloading = 4;
```
tigrannajaryan, line 836: Label this `[Development]`. All new additions to the spec start in this state.

tigrannajaryan (dismissed):

proto/opamp.proto
```diff
@@ -802,6 +802,22 @@ message PackageStatus {
 
     // Error message if the status is erroneous.
     string error_message = 7;
+
+    // Optional details that may be of interest to a user.
+    // Should only be set if status is Downloading.
+    // Status: [Development]
+    PackageDownloadDetails download_details = 8;
```
tigrannajaryan, line 809: Please add this to the specification.md: https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#packagestatus-message

We may need to rethink the need to duplicate this information, but for now it exists in 2 places: in the proto and in the specification.md.

tigrannajaryan (commented):
LGTM

proto/opamp.proto
```diff
@@ -802,6 +802,22 @@ message PackageStatus {
...
+
+
+// Additional details that an agent can use to describe an in-progress package download.
+// Status: [Development]
+message PackageDownloadDetails {
+    // The package download progress as a percentage.
+    double download_percent = 1;
+
+    // The current package download rate in bytes per second.
+    uint64 download_rate = 2;
```
andykellr, line 820: I would prefer to include the unit in the name of the field. I think we could remove rate as it will be clear that it is a rate.

```suggestion
    // The current package download rate in bytes per second.
    uint64 download_bytes_per_second = 2;
```

proto/opamp.proto
michel-laterman: OK, added my changes; also fixed status to include the `INSTALL_PENDING` status that was missing

tigrannajaryan (approved):

specification.md
```diff
@@ -1310,11 +1311,14 @@ message PackageStatus {
     bytes server_offered_hash = 5;
     enum Status {
         INSTALLED = 0;
-        INSTALLING = 1;
-        INSTALL_FAILED = 2;
+        INSTALL_PENDING = 1;
+        INSTALLING = 2;
+        INSTALL_FAILED = 3;
+        DOWNLOADING = 4;
```
andykellr, line 1317: Sorry, just noticed this. Can we keep the existing values the same?
```suggestion
        INSTALLING = 1;
        INSTALL_FAILED = 2;
        INSTALL_PENDING = 3;
        DOWNLOADING = 4;
```

andykellr (changes requested):

specification.md
andykellr: Oh nevermind, this is just the specification.md and the proto is correct.

andykellr (approved):

tigrannajaryan:
@michel-laterman I think this is good to go. It is still shows as a "Draft", but I assume you are done and we can merge?

michel-laterman:
Yes, we can merge at-will. However I'm unable to run `make gen-proto` locally.

tigrannajaryan:
> Yes, we can merge at-will. However I'm unable to run `make gen-proto` locally.

Github actions verified it, we are good: https://github.com/open-telemetry/opamp-spec/actions/runs/11483570522/job/31960049021?pr=206

tigrannajaryan:
Thanks @michel-laterman !

PR #217, opened 2025-03-25
Add spec change for updating capabilities after initial message

jaronoff97:
Following from [this comment](https://github.com/open-telemetry/opamp-go/pull/366#issuecomment-2751917718) from @tigrannajaryan, this PR updates the spec to make it explicitly allowed to update the remote config capacity after the initial capacity.

Please let me know if this is not the right place for this to go, it felt most accurate as there's a similar section in the [servertoagent.capabilities](https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#servertoagentcapabilities) part of the spec:
> This field MUST be set in the first ServerToAgent sent by the Server and MAY be omitted in subsequent ServerToAgent messages by setting it to UnspecifiedServerCapability value.

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -560,6 +560,14 @@ in the future such that old Agents automatically report that they don't
 support the new capability.
 This field MUST be always set.
 
+This field MUST be set in the first AgentToServer sent by the Agent and MAY be
+omitted in subsequent AgentToServer messages by setting it to
+UnspecifiedAgentCapability value. An Agent MAY update its'
+AgentCapabilities.AcceptsRemoteConfig capability at ANY time after the first
+message. If the capability is enabled, the server MAY offer a new remote
+configuration. If the capability is dsiabled, the server MUST NOT send a remote
+configuration update.
+
 ```protobuf
 enum AgentCapabilities {
     // The capabilities field is unspecified.


specification.md
```diff
@@ -560,6 +560,14 @@ in the future such that old Agents automatically report that they don't
 support the new capability.
 This field MUST be always set.
 
+This field MUST be set in the first AgentToServer sent by the Agent and MAY be
+omitted in subsequent AgentToServer messages by setting it to
+UnspecifiedAgentCapability value. An Agent MAY update its'
+AgentCapabilities.AcceptsRemoteConfig capability at ANY time after the first
+message. If the capability is enabled, the server MAY offer a new remote
+configuration. If the capability is dsiabled, the server MUST NOT send a remote
```
tigrannajaryan, line 568:
```suggestion
configuration. If the capability is disabled, the server MUST NOT send a remote
```

specification.md
```diff
@@ -560,6 +560,14 @@ in the future such that old Agents automatically report that they don't
 support the new capability.
 This field MUST be always set.
 
+This field MUST be set in the first AgentToServer sent by the Agent and MAY be
+omitted in subsequent AgentToServer messages by setting it to
+UnspecifiedAgentCapability value. An Agent MAY update its'
+AgentCapabilities.AcceptsRemoteConfig capability at ANY time after the first
```
tigrannajaryan, line 566: Can we make it even more explicit that no other capabilities may change in subsequent messages?

tigrannajaryan:
@andykellr can you please also take a look?

Any suggestions on how to mark this particular change as a new change in [Development] status? Make it a separate section?

tigrannajaryan:
@jaronoff97 can you please also update the prototype at https://github.com/open-telemetry/opamp-go/pull/366 to demonstrate this spec change? We [now require](https://github.com/open-telemetry/opamp-spec?tab=readme-ov-file#proposal-requirements) working prototypes before making spec changes.

jaronoff97:
@tigrannajaryan updated the spec based on the new prototype. I realized that we should probably should be able to set ReportsRemoteConfig as well.

jaronoff97:
@tigrannajaryan I'm not sure I agree. 

1. I don't see a reason why we should require shutting down the connection and then restarting it given we don't do the same for something like `SetAvailableComponents` or `SetFlags` both of which ultimately modify the AgentToServer message – I see the ability to modify AcceptsRemoteConfig in the same way as setting flags or setting available components. It's modifying what the server should know about the agent once the agent has that information. In my use case specifically, restarting the connection from the client may result in the pod going unhealthy and then losing the ability to accept remote configuration effectively sticking the client in a restart loop.
2. I think if the client were to declare it accepted remote config upfront but then refused to apply said remote config, the client would effectively be lying to the server. Although its true that "The Agent may ignore the Remote Configuration offer if it does not want its configuration to be remotely controlled by the Server." In the case of this client, I don't think it should be allowed to accept remote configuration if it has not been elected to do so – IMO it would be an odd experience to propagate a remote configuration to multiple clients with the expectation that only one of them actually applies it, especially if its unclear which should actually be applying it. To me the point of the capabilities list is to convey what the client is capable of doing and those capabilities can indeed change over its lifetime.

tigrannajaryan:
@jaronoff97 can you please clarify your use case? I know you linked to https://github.com/open-telemetry/opentelemetry-operator/issues/3822 but I am not sure I understand why the capability needs to change on the fly.

jaronoff97:
sure, we had a bit of discussion on [slack about it](https://cloud-native.slack.com/archives/C02J58HR58R/p1742935656147189?thread_ts=1742844518.631059&cid=C02J58HR58R), but to summarize: the bridge is a pod running in kubernetes and the remote configuration is essentially editing the YAML of a collector in Kubernetes. If there are two bridge pods running simultaneously and they both try to modify the YAML at the same time we'd result in potentially dangerous scenarios or at a minimum wasted energy. I'm trying to introduce a kubernetes lease via the in-cluster leader election mechanism to determine which pod is allowed to do the remote configuration. This determination happens _after_ a client is already started and may be changed at any point in the lifecycle of a pod.

tigrannajaryan:
I have thought a bit more about this and I think it it is OK to make this change to the spec. In the end the server has to be prepared for the agent to change its capabilities anytime. At the very least this can happen when the agent is upgraded to a newer version that has newer capabilities. Interestingly when plain http transport is used the server may simply observe that the next message from the agent has a different set of capabilities.

So, if the server needs to be ready to process a change in capabilities of an agent I think it is only fair to allow that to also happen while the agent is running and while it keeps the websocket connection open. With the Supervisor model this can happen during agent upgrades too, and the Supervisor will keep the websocket connection open.

Given the above, I don't know if it even makes sense to try to restrict what capabilities may change after the first message. A possible approach is that the spec allows any capabilities to change at any time. The opamp-go implementation may still have restrictions on what capabilities may change without restarting the opamp Client.

tigrannajaryan:
Also discussed this in person with @andykellr @evan-bradley, please add any thoughts you have.

jaronoff97:
@tigrannajaryan should I modify the spec PR to specify that a change may happen at anytime? I can also modify my opamp-go PR to allow the setting of any capability in which case we can re-calculate some of the fields per your concerns [here](https://github.com/open-telemetry/opamp-go/pull/366#issuecomment-2751571192).

specification.md
```diff
@@ -560,6 +560,15 @@ in the future such that old Agents automatically report that they don't
 support the new capability.
 This field MUST be always set.
 
+This field MUST be set in the first AgentToServer sent by the Agent and MAY be
+omitted in subsequent AgentToServer messages by setting it to
+UnspecifiedAgentCapability value. An Agent MAY update its'
```
evan-bradley, line 565:
```suggestion
UnspecifiedAgentCapability value. An Agent MAY update its
```

specification.md
```diff
@@ -560,6 +560,15 @@ in the future such that old Agents automatically report that they don't
 support the new capability.
 This field MUST be always set.
 
+This field MUST be set in the first AgentToServer sent by the Agent and MAY be
```
evan-bradley, line 563: Should we merge these sections somehow? I feel like the second paragraph overrides the first a bit.

specification.md
```diff
@@ -560,6 +560,15 @@ in the future such that old Agents automatically report that they don't
...
+omitted in subsequent AgentToServer messages by setting it to
+UnspecifiedAgentCapability value. An Agent MAY update its'
+AgentCapabilities.AcceptsRemoteConfig OR AgentCapabilities.ReportsRemoteConfig
+capability at ANY time after the first message. If the capability is enabled,
+the server MAY offer a new remote configuration. If the capability is disabled,
+the server MUST NOT send a remote configuration update. Currently, other
+capabilities MUST NOT be updated after the initial message is sent.
```
evan-bradley, line 570: Why do we restrict different capabilities in this regard? Should we make this a "SHOULD NOT"?

jaronoff97:
@tigrannajaryan PTAL at the recent changes. Once approved, I will happily update the opamp-go PR, but I want to wait until an approval is present to ensure I'm not implementing something that will change under me.

evan-bradley:
To recap the offline conversation @tigrannajaryan and I had:

1. Situations like agent upgrades may require agents to report new `AgentCapabilities` messages. Instead of needing to reset the connection to change the agent's capabilities, we should allow the connection to stay alive and update the capabilities on the existing connection.
2. It would be more complex to specify how to fail if the server receives a new `AgentCapabilities` message than to simply require the client and server to be resilient to these changing. The server should update the capabilities it expects a client to support, and clients should simply ignore messages related to capabilities they don't support.

I also don't think there's a reason we need to restrict the capabilities that we allow agents to update.

specification.md
```diff
@@ -560,6 +560,12 @@ in the future such that old Agents automatically report that they don't
 support the new capability.
 This field MUST be always set.
 
+An Agent MAY update any of its capabilities at ANY time after the first
```
tigrannajaryan, line 563:
```suggestion
An Agent MAY update any of its capabilities at any time after the first
```
I think we don't need to capitalize ANY.

tigrannajaryan:
I checked the example server implementation and it only records the agent capabilities on the first status report: https://github.com/open-telemetry/opamp-go/blob/2153cd9788b50e50beda4e4743043a9b837a4d78/internal/examples/server/data/agent.go#L195

So, subsequent capability changes will be ignored by the example server. We will need to update this to work correctly with capability changes.

@andykellr I am curious how does Bindplane server work in this situation. Would it work correctly or it also needs to be updated to correctly process capability changes?

jaronoff97:
I think for the client we'll do a similar approach to how the [SetCustomCapabilities](https://github.com/open-telemetry/opamp-go/blob/main/client/internal/clientstate.go#L153-L166) works today. For the server, it should probably adjust to do the same as we do for `updateAgentDescription`. I'll own updating the opamp-go after the spec change.

tigrannajaryan:
@jaronoff97 the server-side logic that handles the agent capabilities today is in the example server, it is not part of the `Server` code, so I am not sure we can have the equivalent of SetCustomCapabilities. Perhaps I am missing something.

jaronoff97:
i know this, i was just referencing the work needed on the client as well as the work needed on the server example code to make this work effectively

tigrannajaryan:
@jaronoff97 please ping me when https://github.com/open-telemetry/opamp-go/pull/366 is updated to reflect this spec changes.

jaronoff97:
can do, its on my todo list for this week.

PR #220, opened 2025-05-02
Add ConnectionSettingsStatus message and capability

### LINKED ISSUE #211: Clients are unable to indicate when `ConnectionSettingsOffers` fail or succeed
A client is currently unable to signal the server when an attempt to apply settings from a `ConnectionSettingsOffers` message has failed.

There should be a mechanism similar to the client's `RemoteConfigStatus` message it can use to indicate if the settings in `ConnectionSettingsOffers` messages are applied correctly.

michel-laterman:
Add `ConnectionSettingsStatus` that mirrors the form and function of `RemoteConfigStatus` in order to all Clients to report to the Server if the connection settings that are offered have been applied or rejected.

Client use of this new message required the new `AgentCapabilities_ReportsConnectionSettingsStatus` to be set.

- Closes #211

### DIFF
--- a/proto/opamp.proto
+++ b/proto/opamp.proto
@@ -95,6 +95,12 @@ message AgentToServer {
     // A message indicating the components that are available for configuration on the agent.
     // Status: [Development]
     AvailableComponents available_components = 14;
+
+    // The status of the OfferedConnectionSettings that was previously received
+    // from the Server. This field SHOULD be unset if the offered connection
+    // settings status is unchanged since the last AgentToServer message.
+    // Status: [Development]
+    ConnectionSettingsStatus connection_status = 15;
 }
 
 enum AgentToServerFlags {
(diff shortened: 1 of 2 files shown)


michel-laterman:
@andykellr @tigrannajaryan, here is the spec proposal for  reporting connection settings status.
I'll work on the implementation/demo next week.

andykellr:
This looks good, but I'd like to see an implementation to make sure we don't discover any issues in the process.

michel-laterman:
Start of demo/implementation is https://github.com/open-telemetry/opamp-go/pull/390

andykellr (changes requested):
There are a few places where ConnectionStatus is used instead of ConnectionSettingsStatus and I think that is confusing. I also mentioned this in the associated opamp-go PR. We should be consistent and use the full name. ConnectionStatus could mean something different.

I think once this is resolved, I am ready to approve this PR.

proto/opamp.proto
```diff
@@ -700,54 +706,57 @@ enum AgentCapabilities {
...
     // Agent should use its own configured interval, which by default will be 30s. The Server may not
     // know the configured interval and should not make assumptions about it.
     // Status: [Development]
-    AgentCapabilities_ReportsHeartbeat               = 0x00002000;
+    AgentCapabilities_ReportsHeartbeat                = 0x00002000;
     // The agent will report AvailableComponents via the AgentToServer.available_components field.
     // Status: [Development]
-    AgentCapabilities_ReportsAvailableComponents     = 0x00004000;
+    AgentCapabilities_ReportsAvailableComponents      = 0x00004000;
+    // The agent will report ConnectionSettingsOffers status via AgentToServer.connection_settings_status_field.
```
andykellr, line 757:
```suggestion
    // The agent will report ConnectionSettingsOffers status via AgentToServer.connection_settings_status field.
```

proto/opamp.proto
```diff
@@ -798,6 +807,21 @@ message RemoteConfigStatus {
...
+// Status: [Development]
+message ConnectionSettingsStatus {
+    // The hash of the connection settings that was last recieved by this Agent
+    // in the connection_settings.hash field. The Server SHOULD compare this
+    // hash with the OfferedConnectionSettings hash it has for the Agent and if
+    // the hashes are different the Server MUST include the connection_settings
+    // field in the response in the ServerToAgent message.
+    bytes last_connection_settings_hash = 1;
+
+    RemoteConfigStatuses status = 2;
```
tigrannajaryan, line 819: I think it is confusing to have connection setting status reflected via a remote config status enum.

We probably have a couple options:
1. Create a copy of RemoteConfigStatuses enum and call it ConnectionSettingsStatuses.
2. Rename RemoteConfigStatuses to some sort of "Generic" Statuses enum (not sure what's a good name for it). This only works if we are certain the statuses for both of these things are always going to require the same enum set. I haven't thought through it, so not sure if it is a great idea.

tigrannajaryan:
I think this makes sense conceptually, just needs a minor tweak on the enum.

proto/opamp.proto
michel-laterman: @andykellr do you have any opinions on the direction?

I'm leaning towards option 1 as I also can't think of a good rename for option 2.

michel-laterman:
@tigrannajaryan, i'm going to keep the implementation as a draft until the spec change has been released.

andykellr (approved):

PR #243, opened 2025-07-08
Add namespace for C#

RassK:
### What

Defines a C# namespace for generated Protobuf classes.

### DIFF
--- a/proto/anyvalue.proto
+++ b/proto/anyvalue.proto
@@ -23,7 +23,7 @@ syntax = "proto3";
 package opamp.proto.v1;
 
 option go_package = "github.com/open-telemetry/opamp-go/protobufs";
-option csharp_namespace = "OpAmp.Protocol";
+option csharp_namespace = "OpAmp.Protocol.V1";
 
 // AnyValue is used to represent any type of attribute value. AnyValue may contain a
 // primitive value such as a string or integer or it may contain an arbitrary nested
--- a/proto/anyvalue.proto
+++ b/proto/anyvalue.proto
@@ -20,7 +20,7 @@
 
 syntax = "proto3";
 
-package opamp.proto;
+package opamp.proto.v1;
 
 option go_package = "github.com/open-telemetry/opamp-go/protobufs";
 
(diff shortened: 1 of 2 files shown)


RassK:
Could be extended to `OpenTelemetry.OpAmp.Protocol` if OTel branding is also important.

Kielek (approved):

proto/anyvalue.proto
```diff
@@ -23,6 +23,7 @@ syntax = "proto3";
 package opamp.proto;
```
pellared, line 23: _should be as a different issue but I find it very related to this PR_ 

should it not be

```suggestion
package opentelemetry.opamp.proto.v1;
```

to follow our existing patterns in https://github.com/open-telemetry/opentelemetry-proto?

pellared: That about 

```suggestion
package opamp.proto.v1;
```

?

tigrannajaryan: Yeah, that's fine, we can do that.

tigrannajaryan: Unless you need it for C# I suggest we don't touch it in this PR.

proto/anyvalue.proto
```diff
@@ -23,6 +23,7 @@ syntax = "proto3";
...
 
 option go_package = "github.com/open-telemetry/opamp-go/protobufs";
+option csharp_namespace = "OpAmp.Protocol";
```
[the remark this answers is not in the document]
Kielek:
```suggestion
option csharp_namespace = "OpAmp.Protocol.V1";
```

@tigrannajaryan, I think that this option is the best, if we have [agreement for go-lang](https://github.com/open-telemetry/opamp-spec/pull/243#discussion_r2226674570)

@RassK has is on holiday. It will be great, if you can apply suggestion, and merge it.

RassK:
@tigrannajaryan can we get this and #251 reopen - otherwise dotnet repos don't have opamp protos in sync and use manual overrides in the proto files.

proto/anyvalue.proto
```diff
@@ -20,9 +20,10 @@
 
 syntax = "proto3";
 
-package opamp.proto;
+package opamp.proto.v1;
```
tigrannajaryan, line 23: Do you need this change? We need to validate it doesn't break anything or if it does what's the impact.

RassK: Not really that important. I think it will just cause internal renamings of used namespaces, which should be fine.

tigrannajaryan (approved):

PR #251, opened 2025-08-04
Include version in package name

RassK:
### What

Updates package name by adding a version.

ref discussion: https://github.com/open-telemetry/opamp-spec/pull/243#discussion_r2226674570

### DIFF
--- a/proto/anyvalue.proto
+++ b/proto/anyvalue.proto
@@ -20,7 +20,7 @@
 
 syntax = "proto3";
 
-package opamp.proto;
+package opamp.proto.v1;
 
 option go_package = "github.com/open-telemetry/opamp-go/protobufs";
 
--- a/proto/opamp.proto
+++ b/proto/opamp.proto
@@ -16,7 +16,7 @@
 
 syntax = "proto3";
 
-package opamp.proto;
+package opamp.proto.v1;
 
 import "anyvalue.proto";
 


pellared:
AFAIK it can impact the wire format it would be included in field with `google.protobuf.Any` type . See: https://protobuf.dev/programming-guides/proto3/#any.

tigrannajaryan:
> AFAIK it can impact the wire format it would be included in field with `google.protobuf.Any` type . See: https://protobuf.dev/programming-guides/proto3/#any.

But where would we use `google.protobuf.Any`? OpAMP doesn't use it anywhere.

tigrannajaryan:
> Side note: Adding a version suffix is seen as a good practice: https://buf.build/docs/lint/rules/#package_version_suffix

I agree, but only if we are confident the change is safe.

pellared:
>>OpAMP doesn't use it anywhere.


However, others can use it e.g. something like

```proto3
message Event {
  string event_id = 1;
  google.protobuf.Any body = 2;
}
```

Personally, I would but be worried as OpAMP is not stable and this is probably an edge case scenario and  it would even not lead to any compilation errors.

RassK:
@tigrannajaryan can we move forward with this? - that should unblock #243.

tigrannajaryan:
@RassK I don't have a confirmation that this is not breaking anything. I need some proof, a demonstration that this is not a breaking change. Without it I won't move forward with the change.

At the very least I would like somebody to fork, make this change, build opamp-go with this change and confirm that nothing changed in the proto that caused incompatibility with the previous version that doesn't use version in the package name.

@andykellr anything else you would like to be confirmed before we accept the chance?

tigrannajaryan:
@RassK I reopened but my question still stands: https://github.com/open-telemetry/opamp-spec/pull/251#issuecomment-3299237351

tigrannajaryan:
I tested opamp-go examples with server on old proto and client with this modified proto and I don't see any problems, they work fine together.

I tend to agree, we can accept this.

@andykellr @evan-bradley any objections?

evan-bradley:
No objections from me given the results of your tests. Read through the discussion above and I think this makes sense.

evan-bradley (approved):

tigrannajaryan (approved):

tigrannajaryan:
I will keep this open for a bit to give time to @andykellr to review.

andykellr (approved):

PR #306, opened 2026-03-25
Restrict capabilities that may change after the first message

tigrannajaryan:
We made this decision a while back but forgot to update the spec: https://github.com/open-telemetry/opamp-go/pull/366#issuecomment-2751917718

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -589,8 +589,14 @@ in the future such that old Agents automatically report that they don't
 support the new capability.
 This field MUST be always set.
 
-An Agent MAY update any of its capabilities at any time after the first
-message. The server MUST respect the Agent's changes, ensuring that the
+An Agent MAY update (enabled or disable) some of its capabilities at any time after
+the first message. The following capabilities only may be updated after the first
+message:
+
+- AcceptsRemoteConfig
+- ReportsRemoteConfig
+
+The server MUST respect the Agent's capability changes, ensuring that the
 subsequent messages are sent correctly. If the server sends a part of its
 messages corresponding to a capability the Agent doesn't support, the Agent
 SHOULD ignore it.
--- a/.markdownlint.yaml
+++ b/.markdownlint.yaml
@@ -16,6 +16,9 @@ fenced-code-language: false
 # MD022/blanks-around-headings/blanks-around-headers - Headings should be surrounded by blank lines
 MD022: false
 
+# MD024/no-duplicate-heading - Multiple headings with the same content
+MD024: false
+
 # MD025/single-title/single-h1 - Multiple top-level headings in the same document
 MD025: false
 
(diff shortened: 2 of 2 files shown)


tigrannajaryan:
@andykellr @jaronoff97 you probably remember the discussion we had in this thread: https://github.com/open-telemetry/opamp-go/pull/366#issuecomment-2751523548

The last message I posted there was that we will restrict what capabilities may be changed after starting. However, we haven't changed the spec to reflect that.

This PR adds the limitation. Please take a look.

jaronoff97 (approved):

tigrannajaryan:
@evan-bradley I would also like your opinion on this. Does Supervisor ever change the set of supported capabilities after it starts the OpAMP Agent? Can this restriction cause problems for the Supervisor?

tigrannajaryan:
Corresponding change in opamp-go: https://github.com/open-telemetry/opamp-go/pull/538

tigrannajaryan:
@andykellr @evan-bradley please take a look, I would like your opinion on this.

specification.md
```diff
@@ -589,8 +589,14 @@ in the future such that old Agents automatically report that they don't
...
-An Agent MAY update any of its capabilities at any time after the first
-message. The server MUST respect the Agent's changes, ensuring that the
+An Agent MAY update (enabled or disable) some of its capabilities at any time after
+the first message. The following capabilities only may be updated after the first
+message:
+
+- AcceptsRemoteConfig
+- ReportsRemoteConfig
+
+The server MUST respect the Agent's capability changes, ensuring that the
```
andykellr, line 599:
```suggestion
An Agent MAY update (enabled or disable) the following capabilities after the first
message:

- AcceptsRemoteConfig
- ReportsRemoteConfig

The server MUST respect the Agent's capability changes, ensuring that the
```

andykellr (changes requested):
Minor edit for clarity.

evan-bradley:
>Does Supervisor ever change the set of supported capabilities after it starts the OpAMP Agent?

It doesn't right now, since there's no way to make this update; the capabilities can only be toggled through the config file, which is only read on startup.

>Can this restriction cause problems for the Supervisor?

I don't think so, no. Neither of these capabilities are present in the Collector, which is where any issues would arise, since the Collector's capabilities are expected to remain the same over its lifetime.

tigrannajaryan:
I have put some more thought into this and I do not think this is the right approach. There is no way to enforce this agent's behavior for the server, so the server has to be prepared for capability changes anyway.

I think the right approach is to restrict allowed capability changes in the client implementations, limiting them to only tested permutations. As new scenarios require other capability changes to be allowed implementations should work on supporting them, with corresponding testing verifying the implementation done.

@andykellr @evan-bradley I am closing this. Let me know if you disagree with the above and we can discuss further.

PR #310, opened 2026-04-06
Clarify what language agents should use for identity

breedx-splk:
The original opamp spec was written with the Collector at top of mind. 

There is increasing interest in using OpAMP to manage or re/configure certain aspects of language agents at runtime. This PR adds some clarification about how language instrumentation agents should handle identity and non-identity attributes in OpAMP...effectively just passing along the Resource and nothing more.


aunshc:
Raised this topic at the SIG, the recommendation from the maintainers was to keep the spec generic and create a supplemental document for agent (eg. language instrumentation) specific aspects. @tigrannajaryan @andykellr we'd probably want to move the existing standalone section in context to the Collector out as well in that case?

specification.md
```diff
@@ -1221,8 +1227,19 @@ telemetry. The combination of identifying attributes SHOULD be sufficient to
 uniquely identify the Agent's own telemetry in the destination system to which
 the Agent sends its own telemetry.
 
+###### Language Instrumentation Agents
+
+Instrumentation agents use an OpenTelemetry SDK instance which includes a
```
RassK, line 1232: why not also filter `service.name`, `service.namespace`, `service.version`, `service.instance.id` to `identifying_attributes` and everything else to `nonidentifying_attributes` ?

specification.md
```diff
@@ -1204,13 +1208,15 @@ Attributes that identify the Agent.
 Keys/values are according to OpenTelemetry [resource semantic
 conventions](https://opentelemetry.io/docs/specs/semconv/resource/).
 
-For standalone running Agents (such as OpenTelemetry Collector) the following
-attributes SHOULD be specified:
+###### Standalone Agents
```
tigrannajaryan, line 1211: @breedx-splk let's move this section and the new "Language Instrumentation Agents" section to a separate document. We can call it something like `opentelemetry-guidelines.md` and it will contain all considerations that are specific to Otel agents. The `specification.md` will contain generic protocol specification, which is 99% of the content right now (the only exception is this paragraph, so let's move it out).

specification.md
RassK: CC: @stevejgordon

breedx-splk:
💨

PR #312, opened 2026-04-07
Add OpenTelemetry guidelines doc

breedx-splk:
This follows #310.

It creates a new document instead of polluting the general high level OpAMP spec with tedious details about actual behavior.

### DIFF
--- a/opentelemetry-guidelines.md
+++ b/opentelemetry-guidelines.md
@@ -1,0 +1,33 @@
+# OpenTelemetry Guidelines
+
+This document provides guidance for language instrumentation agents that use
+OpAMP.
+
+## Scope
+
+This guidance is intended for OpenTelemetry language agents that run as part of an instrumented
+application process. Examples include auto-instrumentation language agents and similar
+instrumentation runtimes built with OpenTelemetry SDKs.
+
+This guidance does not apply to standalone agents such as the OpenTelemetry
+Collector.
+
+## Identifying Attributes
+
+Language instrumentation agents SHOULD copy all attributes from the
+OpenTelemetry SDK `Resource` into `AgentDescription.identifying_attributes`.
+
+Additional identifying attributes MAY be included if they are needed to
+uniquely identify the agent in the deployment environment.
+
+## Non-identifying Attributes
+
+Language instrumentation agents SHOULD leave
+`AgentDescription.non_identifying_attributes` empty.
+
+## Rationale
+
+For language instrumentation agents, the OpenTelemetry SDK `Resource` already
+defines the identity of the telemetry source. Reusing those attributes in OpAMP
+keeps the agent identity aligned with the identity used by the telemetry that
+the agent produces.
--- a/opentelemetry-guidelines.md
+++ b/opentelemetry-guidelines.md
@@ -14,20 +14,26 @@ Collector.
 
 ## Identifying Attributes
 
-Language instrumentation agents SHOULD copy all attributes from the
-OpenTelemetry SDK `Resource` into `AgentDescription.identifying_attributes`.
+Language instrumentation agents MUST copy the following OpenTelemetry SDK
+`Resource` attributes into `AgentDescription.identifying_attributes`:
 
-Additional identifying attributes MAY be included if they are needed to
-uniquely identify the agent in the deployment environment.
+- `service.name`
+- `service.instance.id`
+- `service.namespace.name`, if present
+
+These identifying attributes MUST match the values that the agent uses in the
+Resource of its own telemetry.
 
 ## Non-identifying Attributes
 
-Language instrumentation agents SHOULD leave
-`AgentDescription.non_identifying_attributes` empty.
+Language instrumentation agents SHOULD copy all other OpenTelemetry SDK
+`Resource` attributes into `AgentDescription.non_identifying_attributes`,
+excluding those added to the identifying attributes (above).
 
 ## Rationale
 
-For language instrumentation agents, the OpenTelemetry SDK `Resource` already
-defines the identity of the telemetry source. Reusing those attributes in OpAMP
-keeps the agent identity aligned with the identity used by the telemetry that
-the agent produces.
+For language instrumentation agents, the `service.*` attributes listed above
+define the identity that OpAMP uses to associate an agent with its telemetry.
+Copying the remaining Resource attributes into
+`AgentDescription.non_identifying_attributes` preserves useful descriptive
+context without expanding the agent identity beyond those service attributes.
--- a/README.md
+++ b/README.md
@@ -4,6 +4,8 @@ This repository contains Open Agent Management Protocol (OpAMP)
 [specification](specification.md) and Protobuf definitions in [proto](proto) directory.
 
 See [releases here](https://github.com/open-telemetry/opamp-spec/releases) and [changelog here](CHANGELOG.md).
+Additional standalone guidance for instrumentation runtimes is in
+[opentelemetry-guidelines.md](opentelemetry-guidelines.md).
 
 ## Contributing
 


tigrannajaryan (approved):

tigrannajaryan:
@open-telemetry/specs-approvers @open-telemetry/technical-committee please review. 

This is opinionated take on how Otel SDKs should identify themselves via OpAMP in a remotely managed system.

opentelemetry-guidelines.md
```diff
@@ -0,0 +1,33 @@
...
+application process. Examples include auto-instrumentation language agents and similar
+instrumentation runtimes built with OpenTelemetry SDKs.
+
+This guidance does not apply to standalone agents such as the OpenTelemetry
+Collector.
+
+## Identifying Attributes
+
+Language instrumentation agents SHOULD copy all attributes from the
+OpenTelemetry SDK `Resource` into `AgentDescription.identifying_attributes`.
```
RassK: https://github.com/open-telemetry/opamp-spec/pull/310#discussion_r3044727858

I do not get the idea for a single exception here, while everything else follows a different pattern?
https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#agentdescriptionidentifying_attributes

RassK: why should a `os.version` be in the `identifying_attributes` list while here it's said otherwise:
https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#agentdescriptionnon_identifying_attributes

breedx-splk: @tigrannajaryan can probably help clarify this, but I think the idea is that language agents and the Collector treat Resource and telemetry slightly differently. Notably, the Collector's Resource for its own internal/diagnostic telemetry is often different than the Resource(s) of the things it is handling telemetry for (like other agents or the operating system or whatever)...whereas the language agents use a Resource to identify the source of ALL telemetry it produces.

RassK: In the end it's "2 pipes taking to the same place" anyway, so server can rearrange however it needs - but the change itself will be a breaking change if there is a need for one.

I just wanted to be sure that the base hypothesis about what the resource contains is same. At least in .NET it contains tons of general classifiers like `os.version`, `host.arch`, `process.runtime.name`.

@tigrannajaryan said that `identifying_attributes` should form a guid but in the end every extra component makes that guid stronger (besides being a very dynamic guid) 🤷‍♂️

tigrannajaryan: > [#310 (comment)](https://github.com/open-telemetry/opamp-spec/pull/310#discussion_r3044727858)
> 
> I do not get the idea for a single exception here, while everything else follows a different pattern? https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#agentdescriptionidentifying_attributes

We should copy that section into this file too. Those are also OpenTelemetry guidelines, for the Collector.

tigrannajaryan: > why should a `os.version` be in the `identifying_attributes` list while here it's said otherwise: https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#agentdescriptionnon_identifying_attributes

There is no contradiction here. `os.version` is not in Collector's Resource, thus we do not expect it to be in `identifying_attributes`. But since information about OS used it typically useful to agents the guidelines recommend including it as `non_identifying_attributes`.

tigrannajaryan: >I just wanted to be sure that the base hypothesis about what the resource contains is same. At least in .NET it contains tons of general classifiers like os.version, host.arch, process.runtime.name.

This is the unfortunate result of how we are forced to put everything into Resource which is just a single bag of attributes. We simply have no way to differentiate attributes that are important for identification vs ones that are useful to have but are not required for identification. The new [Entity concept](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/entities/data-model.md) solves this problem by having this clear distinction between identifying and non-identifying attributes, which mirror OpAMP concepts precisely. 

For now I think it is fine for simplicity to say we include all Resource attributes.

Kielek (approved):

opentelemetry-guidelines.md
RassK: Thanks for confirming!

RassK (approved):
> > [#310 (comment)](https://github.com/open-telemetry/opamp-spec/pull/310#discussion_r3044727858)
> > I do not get the idea for a single exception here, while everything else follows a different pattern? https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#agentdescriptionidentifying_attributes
> 
> We should copy that section into this file too. Those are also OpenTelemetry guidelines, for the Collector.

With this minor change, it's seems good to go.

tigrannajaryan:
I am going to keep this open for a few days to get more extensive feedback from the community. I will publicize in Slack.

jack-berg (approved):

dashpole (approved):

opentelemetry-guidelines.md
```diff
@@ -0,0 +1,33 @@
+# OpenTelemetry Guidelines
+
+This document provides guidance for language instrumentation agents that use
+OpAMP.
```
pellared, line 4: (nit) Can we settle on using a single term? Previously we have

> guidance for instrumentation runtimes

Personally, I am not a fan of using the "agent" term. I suggest "component" or "runtime" or "implementation" (I do not have any strong opinion).  I think the sentence below is the only one that where using "agent" is fine (however, we could also use zero-code instrumentation to align with https://opentelemetry.io/)

> Examples include auto-instrumentation language agents and similar
instrumentation runtimes built with OpenTelemetry SDKs.

What matters for me the most is having consistent terminology used in this PR.

tigrannajaryan:
@dmitryax please take a look.

andykellr (approved):

opentelemetry-guidelines.md
```diff
@@ -0,0 +1,33 @@
...
+
+Additional identifying attributes MAY be included if they are needed to
+uniquely identify the agent in the deployment environment.
+
+## Non-identifying Attributes
+
+Language instrumentation agents SHOULD leave
+`AgentDescription.non_identifying_attributes` empty.
+
+## Rationale
```
evan-bradley, line 33:
```suggestion
### Rationale
```
Mostly a nit: not sure whether this section or the identifying attributes section would be better, but shouldn't this be nested under one of those two? I don't think this rationale necessarily applies to the whole document.

opentelemetry-guidelines.md
```diff
@@ -0,0 +1,33 @@
...
+
+For language instrumentation agents, the OpenTelemetry SDK `Resource` already
+defines the identity of the telemetry source. Reusing those attributes in OpAMP
+keeps the agent identity aligned with the identity used by the telemetry that
+the agent produces.
```
dmitryax: The rationale assumes that the SDK's Resource is a usable identity, but recent entities work specifically distinguishes identifying vs. descriptive attributes within Resource because:

1. **Descriptive attributes are mutable** and can change during an entity's lifetime without changing its identity. Copying all of Resource into `identifying_attributes` would cause spurious agent identity churn on every such change. This is modeled in [OTEP 0264](https://github.com/open-telemetry/opentelemetry-specification/blob/main/oteps/entities/0264-resource-and-entities.md) (Attempt to bring it to the SDK level: open-telemetry/opentelemetry-specification#4565).

2. **Descriptive attributes can contain complex values** after open-telemetry/opentelemetry-specification#4485 is merged which should not be used as identity.  The entity spec intentionally restricts `entity.id` to `map<string, string>` while `entity.description` uses `map<string, AnyValue>`.

3. **Resource is a composite of multiple entities**, and the set grows as detectors are added. For example, open-telemetry/opentelemetry-specification#4768 defines an explicitly additive model: each detector appends new entity types without affecting existing ones. This means enabling a k8s or cloud detector adds new entities to the resource but does not change the agent's identity. Treating all resource attributes as `identifying_attributes` inverts this, every new detector would produce a new OpAMP agent identity for the same running process.

Why can't we use the set identifying attributes of entities that are always emitted by the SDKs (`service.name`, `service.instance.id`, `service.namespace.name`?) as a basis for the set of identifying attributes emitted with OpAMP and everything else would be set to the non-identifying attributes?

tigrannajaryan: >This means enabling a k8s or cloud detector adds new entities to the resource but does not change the agent's identity.

@dmitryax It is good to be able to preserve identity unchanged when detectors are enabled/disabled, although I am not sure generally speaking this should be the expectation. If you change configuration of agent you can't expect that the telemetry or other status (e.g. opamp identity) it reports remains unchanged.

>Why can't we use the set identifying attributes of entities that are always emitted by the SDKs (service.name, service.instance.id, service.namespace.name?) as a basis for the set of identifying attributes emitted with OpAMP and everything else would be set to the non-identifying attributes?

This is a possible option that I have thought about. We are guaranteeing that this triplet is globally unique. @breedx-splk thoughts?

tigrannajaryan (dismissed):
All, I have been thinking about [the suggestion](https://github.com/open-telemetry/opamp-spec/pull/312#discussion_r3090215433) @dmitryax made and I think it is a good one.

We can hard code the triplet (service.name, service.instance.id, service.namespace.name) to be the identifying attributes for Otel SDKs. 

All other Resource attributes can go into non-identifying attributes. As Dmitrii said if you enable/disable detectors or the environment changes these other attributes can change, which is not desirable for identifying attributes.

This matches the 2 formal requirements OpAMP spec has for identifying attributes, quoting from the spec:
>The Agent SHOULD also include these attributes in the Resource of its own telemetry.

Yes, these attributes will be in own telemetry (if SDK reports own telemetry).

>The combination of identifying attributes SHOULD be sufficient to uniquely identify the Agent's own telemetry in the destination system to which the Agent sends its own telemetry.

True again, the uniqueness is guaranteed. We [already guarantee](https://opentelemetry.io/docs/specs/semconv/resource/service/#service-instance) that the triplet is globally unique. 

This also nicely aligns conceptually with how the upcoming Entities think about identifying and non-identifying attributes. The triplet is precisely the identifying attributes of the Service Instance entity.

@breedx-splk I know we earlier agreed the full Resource is the right choice, but I think I was wrong, this is a better approach.

Blocking this to prevent merging for now until we discuss this option.

opentelemetry-guidelines.md
breedx-splk: It's indeed a nit, and I don't feel that strongly about it. With whatever term you go with, you'll have people wondering about the other term. Using both helps to suggest that both are valid, but whatever, I get it....and I agree that "agent" is criminally overloaded. Not sure which term is the preferred term of art anymore honestly....

breedx-splk:
I should point out that @RassK suggested essentially the same thing over here in the original PR that I closed: https://github.com/open-telemetry/opamp-spec/pull/310#discussion_r3044727858 

I honestly do not care much about this stuff, so I don't have a real opinion. I will change this PR to reflect the new guidance: `service.name`, `service.instance.id`, `service.namespace.name` into identifying, everything else from the resource into non-identifying. 👍🏻

breedx-splk:
Ok, I did the thing. I wonder if omitting `deployment.environment.name` is an oversight....but I'm not sure. I thought that users used `deployment.environment.name` to segregate environments (like staging, pre-prod, prod, us-west, etc)...but really, the instance id is probably enough, so whatever. 🤷🏻

tigrannajaryan:
> Ok, I did the thing. I wonder if omitting `deployment.environment.name` is an oversight....but I'm not sure. I thought that users used `deployment.environment.name` to segregate environments (like staging, pre-prod, prod, us-west, etc)...but really, the instance id is probably enough, so whatever. 🤷🏻

Thanks.

Yes, we can still use `deployment.environment.name`, which will go into non-identifying attributes. I think this is fine, it will be available in the backend for e.g. supplying different configs to different environments, but it is not necessary for globally unique identification, so we don't need it in the identifying attributes.

tigrannajaryan:
@andykellr @dashpole @jack-berg @RassK @Kielek this PR has changed substantially. Please take another look.

dmitryax (approved):

opentelemetry-guidelines.md
```diff
@@ -0,0 +1,39 @@
...
+This guidance does not apply to standalone agents such as the OpenTelemetry
+Collector.
+
+## Identifying Attributes
+
+Language instrumentation agents MUST copy the following OpenTelemetry SDK
+`Resource` attributes into `AgentDescription.identifying_attributes`:
+
+- `service.name`
+- `service.instance.id`
```
Kielek, line 21: Is it intentional to be here?
If not provided manually, it is generated during startup process. At least in the OTel .NET Auto.
After the restart, it will provide new id.

I am pretty sure that it was common behavior in other technologies at the time of implementation.

Ref: https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/blob/1ad703dec2062ebc50cef55dd76c81b8e8c08b11/src/OpenTelemetry.AutoInstrumentation/Configurations/ResourceConfigurator.cs#L51-L55

Kielek (approved):
It should be also fine, with some comment related to `service.instance.id`

opentelemetry-guidelines.md
breedx-splk: I didn't suggest it, but I understood it to be intentional, yes. 

Without it, you're just left with `service.name` and `service.namespace`, which is too ambiguous and doesn't serve to uniquely identify an agent instance.

andykellr (approved):

tigrannajaryan:
Merging. If further refinements are needed, can be done in a future PR.

tigrannajaryan:
Continuation which moves Otel Collector guidelines to the same doc: https://github.com/open-telemetry/opamp-spec/pull/318

PR #314, opened 2026-04-15
Add Duplicate instance_uid detection section

### LINKED ISSUE #20: Do we need to protect against Agent instance ID reuse / impersonation?
How does the server know that the request/report received actually is generated by the agent that the message claims? Do we need to verify that the message is not fake (impersonated) or the server trusts the senders that are authenticated?

michel-laterman:
As discussed in OpAMP SIG meeting (April 14, 2026), we are adding a section that applies the same approach as duplicate websocket detection to duplicate `instance_uid` detection.
I've added a brief description for motive, but have left to duplicate detection mechanism as an implementation specific detail

- Closes: https://github.com/open-telemetry/opamp-spec/issues/20

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -213,6 +213,7 @@ Status: [Beta]
     + [Plain HTTP Transport](#plain-http-transport-1)
   * [Restoring WebSocket Connection](#restoring-websocket-connection)
   * [Duplicate WebSocket Connections](#duplicate-websocket-connections)
+    + [Duplicate instance_uid Detection](#duplicate-instance_uid-detection)
   * [Authentication](#authentication)
   * [Bad Request](#bad-request)
   * [Retrying Messages](#retrying-messages)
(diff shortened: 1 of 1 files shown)


michel-laterman:
@tigrannajaryan @andykellr, can you please take a look?

specification.md
```diff
@@ -213,6 +213,7 @@ Status: [Beta]
     + [Plain HTTP Transport](#plain-http-transport-1)
   * [Restoring WebSocket Connection](#restoring-websocket-connection)
   * [Duplicate WebSocket Connections](#duplicate-websocket-connections)
+    + [Duplicate instance_uid Detection](#duplicate-instance_uid-detection)
```
andykellr: This language looks useful to me, but did you mean to add it as a subsection of Duplicate WebSocket Connections? During the discussion in the SIG meeting you mentioned that this should apply to HTTP as well.

michel-laterman: Oops, I'll set it as it's own section

andykellr (approved):

PR #318, opened 2026-04-23
Move Otel Collector guidelines from to opentelemetry-guidelines.md

tigrannajaryan:
Continuation of https://github.com/open-telemetry/opamp-spec/pull/312

This brings Otel Collector guidelines from specification.md to the opentelemetry-guidelines.md where we want all Otel-related guidelines to be.

### DIFF
--- a/opentelemetry-guidelines.md
+++ b/opentelemetry-guidelines.md
@@ -1,18 +1,47 @@
 # OpenTelemetry Guidelines
 
-This document provides guidance for language instrumentation agents that use
-OpAMP.
+This document provides guidance for OpenTelemetry agents that use OpAMP.
 
-## Scope
+## Collector
+
+This guidance is intended for OpenTelemetry Collector.
+
+### Identifying Attributes
+
+The OpenTelemetry Collector SHOULD specify the following attributes in
+`AgentDescription.identifying_attributes`:
+
+- `service.name` should be set to the same value that the Agent uses in its own telemetry.
+- `service.namespace` if it is used in the environment where the Agent runs.
+- `service.version` should be set to version number of the Agent build.
+- `service.instance.id` should be set. It may be set equal to the Agent's
+  instance uid (equal to ServerToAgent.instance_uid field) or any other value
+  that uniquely identifies the Agent in combination with other attributes.
+- any other attributes that are necessary for uniquely identifying the Agent's
+  own telemetry.
+
+These values SHOULD match the values that the Collector uses in the Resource of its
+own telemetry.
+
+### Non-identifying Attributes
+
+The following attributes SHOULD be included in `AgentDescription.non_identifying_attributes`:
+
+- `os.type`, `os.version` - to describe where the Collector runs.
+- `host.\*` to describe the host the Collector runs on.
+- `cloud.\*` to describe the cloud where the host is located.
+- any other relevant Resource attributes that describe this Collector and the
+  environment it runs in.
+- any user-defined attributes that the end user would like to associate with
+  this Collector.
+
+## SDKs
 
 This guidance is intended for OpenTelemetry language agents that run as part of an instrumented
 application process. Examples include auto-instrumentation language agents and similar
 instrumentation runtimes built with OpenTelemetry SDKs.
 
-This guidance does not apply to standalone agents such as the OpenTelemetry
-Collector.
-
-## Identifying Attributes
+### Identifying Attributes
 
 Language instrumentation agents MUST copy the following OpenTelemetry SDK
 `Resource` attributes into `AgentDescription.identifying_attributes`:
@@ -24,7 +53,7 @@ Language instrumentation agents MUST copy the following OpenTelemetry SDK
 These identifying attributes MUST match the values that the agent uses in the
 Resource of its own telemetry.
 
-## Non-identifying Attributes
+### Non-identifying Attributes
 
 Language instrumentation agents SHOULD copy all other OpenTelemetry SDK
 `Resource` attributes into `AgentDescription.non_identifying_attributes`,
--- a/specification.md
+++ b/specification.md
@@ -1204,18 +1204,6 @@ Attributes that identify the Agent.
 Keys/values are according to OpenTelemetry [resource semantic
 conventions](https://opentelemetry.io/docs/specs/semconv/resource/).
 
-For standalone running Agents (such as OpenTelemetry Collector) the following
-attributes SHOULD be specified:
-
-- service.name should be set to the same value that the Agent uses in its own telemetry.
-- service.namespace if it is used in the environment where the Agent runs.
-- service.version should be set to version number of the Agent build.
-- service.instance.id should be set. It may be set equal to the Agent's
-  instance uid (equal to ServerToAgent.instance_uid field) or any other value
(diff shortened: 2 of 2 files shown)


tigrannajaryan:
@open-telemetry/collector-approvers please review.

specification.md
```diff
@@ -1226,9 +1214,9 @@ the Agent sends its own telemetry.
...
 runs.
 
-The following attributes SHOULD be included:
+Here are some examples of attributes that may be a good fit for non-identifying attributes:
 
-- os.type, os.version - to describe where the Agent runs.
+- os.\* to describe where the Agent runs.
 - host.\* to describe the host the Agent runs on.
 - cloud.\* to describe the cloud where the host is located.
 - any other relevant Resource attributes that describe this Agent and the
```
tigrannajaryan, line 1222: I kept these as generic recommendations. There is also slightly specific requirements for the Collector in opentelemetry-guidelines.md.

I am not sure how to best differentiate the too, but I felt like some generic recommendations are still useful to have in this file.

andykellr (approved):

opentelemetry-guidelines.md
```diff
@@ -1,18 +1,44 @@
...
+`AgentDescription.identifying_attributes`:
+
+- service.name should be set to the same value that the Agent uses in its own telemetry.
+- service.namespace if it is used in the environment where the Agent runs.
+- service.version should be set to version number of the Agent build.
+- service.instance.id should be set. It may be set equal to the Agent's
+  instance uid (equal to ServerToAgent.instance_uid field) or any other value
+  that uniquely identifies the Agent in combination with other attributes.
+- any other attributes that are necessary for uniquely identifying the Agent's
+  own telemetry.
```
jade-guiton-dd: I suggest we copy something similar to _"The combination of identifying attributes SHOULD be sufficient to uniquely identify the Agent's own telemetry in the destination system to which the Agent sends its own telemetry"_ in this document as well.

There is already something similar for `service.name`, but I'm especially worried about `service.instance.id`, which as-is is very easy to interpret as "put in whatever random instance UID the OpAMP extension decided on, regardless of the Collector's `service.instance.id`".

Honestly, I'd argue the wording for SDKs is clearer and stronger: _"These identifying attributes MUST match the values that the agent uses in the Resource of its own telemetry."_ Maybe it can be reused for the Collector as well.

opentelemetry-guidelines.md
```diff
@@ -1,18 +1,44 @@
...
+
+### Identifying Attributes
+
+The OpenTelemetry Collector SHOULD specify the following attributes in
+`AgentDescription.identifying_attributes`:
+
+- service.name should be set to the same value that the Agent uses in its own telemetry.
+- service.namespace if it is used in the environment where the Agent runs.
+- service.version should be set to version number of the Agent build.
+- service.instance.id should be set. It may be set equal to the Agent's
```
codeboten:
```suggestion
- `service.name` should be set to the same value that the Agent uses in its own telemetry.
- `service.namespace` if it is used in the environment where the Agent runs.
- `service.version` should be set to version number of the Agent build.
- `service.instance.id` should be set. It may be set equal to the Agent's
```

opentelemetry-guidelines.md
```diff
@@ -1,18 +1,44 @@
...
+- any other attributes that are necessary for uniquely identifying the Agent's
+  own telemetry.
+
+### Non-identifying Attributes
+
+The following attributes SHOULD be included in `AgentDescription.non_identifying_attributes`:
+
+- os.type, os.version - to describe where the Collector runs.
+- host.\* to describe the host the Collector runs on.
+- cloud.\* to describe the cloud where the host is located.
```
codeboten:
```suggestion
- `os.type`, `os.version` - to describe where the Collector runs.
- `host.\*` to describe the host the Collector runs on.
- `cloud.\*` to describe the cloud where the host is located.
```

codeboten (approved):

opentelemetry-guidelines.md
tigrannajaryan: Added "These values SHOULD match the values that the Collector uses in the Resource of its 
own telemetry.", which mirrors the spec wording.

opentelemetry-guidelines.md
tigrannajaryan: Fixed.

opentelemetry-guidelines.md
tigrannajaryan: Fixed.

opentelemetry-guidelines.md
```diff
@@ -1,18 +1,47 @@
...
+`AgentDescription.identifying_attributes`:
+
+- `service.name` should be set to the same value that the Agent uses in its own telemetry.
+- `service.namespace` if it is used in the environment where the Agent runs.
+- `service.version` should be set to version number of the Agent build.
+- `service.instance.id` should be set. It may be set equal to the Agent's
+  instance uid (equal to ServerToAgent.instance_uid field) or any other value
+  that uniquely identifies the Agent in combination with other attributes.
+- any other attributes that are necessary for uniquely identifying the Agent's
+  own telemetry.
```
dmitryax, line 21: Do we still need this given https://github.com/open-telemetry/opamp-spec/pull/312?

opentelemetry-guidelines.md
```diff
@@ -1,18 +1,47 @@
...
+This guidance is intended for OpenTelemetry Collector.
+
+### Identifying Attributes
+
+The OpenTelemetry Collector SHOULD specify the following attributes in
+`AgentDescription.identifying_attributes`:
+
+- `service.name` should be set to the same value that the Agent uses in its own telemetry.
+- `service.namespace` if it is used in the environment where the Agent runs.
+- `service.version` should be set to version number of the Agent build.
```
dmitryax, line 16: This also goes against https://github.com/open-telemetry/opamp-spec/pull/312... Shouldn't this be non-identifying attribute?

tigrannajaryan: I am not sure I follow. Why is this against #312?

dmitryax: `service.version` isn't listed in [the set of required identifying attributes](https://github.com/open-telemetry/opamp-spec/pull/312/changes#diff-61b463bb16e4b8aa9b601a413980e342494150b9a0083aed4e8d349ca3734d41R20-R22). Why do we recommend to set it there for the collector?

tigrannajaryan: #312 is for language agents (SDKs). This section is for Collector. For Collector we already have `service.version` listed as required in the [specification.md](https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#agentdescriptionidentifying_attributes)

This PR does not change that requirement. It merely moves existing requirement from `specification.md` to `opentelemetry-guidelines.md`.

We can discuss separately if we want to remove `service.version`, but probably best to do it in a separate PR.

dmitryax: Sounds good

dmitryax (approved):

opentelemetry-guidelines.md
tigrannajaryan: Added https://github.com/open-telemetry/opamp-spec/issues/320

PR #329, opened 2026-05-07
Add attributes to AgentConfigFile

### LINKED ISSUE #328: Add attributes to AgentConfigFile
OpAMP remote configuration currently models configuration as a map of config file names to file bodies. This works well for delivering bytes, but it leaves no structured place to attach metadata about each configuration item. As a result, agents and servers that need identity, ownership, hints for the server, validation context, or backend-specific targeting must encode that information into file names, file bodies, or out-of-band conventions.

I propose adding an optional attributes field to OpAMP config map entries, likely on AgentConfigFile, using a simple string map or existing key/value attribute type.

```
message AgentConfigFile {
  bytes body = 1;
  string content_type = 2;
  map<string, string> attributes = 3;
}
```

Attributes would be optional and advisory. Existing clients and servers can ignore them, preserving backward compatibility. Systems that understand specific attributes can use them to make configuration delivery more explicit and less dependent on custom schemas.

Use cases include:

- We can specify schema versions, merge strategies, or immutability hints outside of the config itself. This keeps the body "pure" so if an agent expects a raw YAML or JSON file, it doesn't have to parse out protocol-specific wrappers first.
- For UI and auditing, we can see exactly what a config is for (or where it came from) at a glance. We shouldn't have to peek into the raw bytes of every config document just to identify its purpose in a dashboard.
- It gives us a clean way to attach ownership info (like which controller or system generated the config) without having to prefix filenames or append "generated by" comments to the actual byte body.
- Servers and agents can use these attributes to quickly decide if a config is compatible or requires specific capabilities before they even attempt to apply the bytes.
- Allows bridge agents or proxies that manage configurations for multiple downstream resources. Each entry can carry its own unique identity and source metadata independently.
- Allows agents with multiple configurations to push each one with identifying metadata.

owais:
Adds optional per-config-file key/value attributes so agents can report arbitrary metadata alongside each config body and content type.

Fixes #328

### DIFF
--- a/proto/opamp.proto
+++ b/proto/opamp.proto
@@ -1046,6 +1046,11 @@ message AgentConfigFile {
     // Optional MIME Content-Type that describes what's in the body field, for
     // example "text/yaml".
     string content_type = 2;
+
+    // Optional key/value pairs that provide additional metadata for this config
+    // file or section. Agents may use this field to report arbitrary attributes
+    // associated with the file.
+    repeated KeyValue attributes = 3;
 }
 
 /////////////////////////////////////////////////////////////////////////////////////
--- a/specification.md
+++ b/specification.md
@@ -2554,6 +2554,7 @@ following structure:
 message AgentConfigFile {
   bytes body = 1;
   string content_type = 2;
+  repeated KeyValue attributes = 3;
 }
 ```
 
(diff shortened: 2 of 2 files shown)


tigrannajaryan (changes requested):
@owais and I discussed this in a call. Blocking for now while he is exploring an alternate option.

RassK:
> @owais and I discussed this in a call. Blocking for now while he is exploring an alternate option.

Any details why an alternative approach is needed? I was also thinking about the same approach. Vendor's spec can define the actual attributes while this spec is dynamic for everything.

tigrannajaryan:
> Any details why an alternative approach is needed?

Alternative does not require any changes to opamp spec. It works with opamp as is, as it was intended.

owais:
@RassK I was able to change my implementation so that it doesn't depend on this request anymore but curious to learn about your use case.

tigrannajaryan:
Closing as not needed anymore.

PR #321, opened 2026-04-30
OpAMP Vision and Roadmap

aunshc:
This PR adds a Vision and Roadmap document for OpAMP that captures:
- Guiding principles
- Use Cases Out of Scope
- Goals identified by OpAMP SIG mantainers

Reference: https://docs.google.com/document/d/1DXXAZmm_pG2ls1zZTzXzUI9bBscunBmcpVbbbN267og/edit?tab=t.0#heading=h.20ds8k5potm

The next step is use milestones in GitHub to group issues from the document for each particular goal.

PR #333, opened 2026-05-12
Add Message Attestation for ServerToAgent messages

### LINKED ISSUE #265: OpAMP Server is treated as trusted root authority for messages, resulting in potential vulnerabilities
The OpAMP Server is currently treated as both the distribution point for ServerToAgent messages as well as the root authoritative source for the same. These include varying categories of messages that, should an OpAMP Server be compromised, could cause unintended behavior, data loss, or in worst case, a breach of sensitive information or takeover of host machines.

### Background

Existing models for distributing packages/updates already have a separate concept of “secure distribution server” and “trusted root authority.” These include, but are not limited to:

* Google Play Store \- distribution server separate from root authority ([Google Play signed package](https://support.google.com/googleplay/android-developer/answer/9842756?hl=en))  
* [Windows System updates](https://learn.microsoft.com/en-us/windows/deployment/update/windows-update-security) \- downloaded via TLS-secured endpoints but are signed/verified by the publisher/vendor  
* [RPM package repositories](https://www.redhat.com/en/blog/secure-distribution-rpm-packages) \- secure connection to the distribution source happens via TLS/HTTPS but the packages themselves are signed by GPG and metadata verified via SHA256 checksum

All of these models use a secure endpoint as a “distribution server” but also have additional separation of concern for a “root authority” of the message, signed by a trusted root/vendor. From hereout I will refer to these concerns as “distribution server” and “trusted root” or “trusted root authority” for clarity.

### Attack Vectors/Vulnerabilities

For example, if an OpAMP Server were to be compromised, the following types of attacks could be executed against an agent that accepts remote config and restart commands:

1. Forced restarts/DOS  
   1. OpAMP Server repeatedly sends ServerToAgentCommand.Restart to the OpAMP Agent.   
   2. Agent will accept these commands believing they are coming from a trusted root (the OpAMP Server), and restart repeatedly, causing a loss of telemetry/outage  
2. Telemetry redirection/”stealing” data  
   1. Server sends AgentRemoteConfig with undesired fields (i.e. changing OTLP Exporter endpoint to a different address, e.g. [vendor1.com/v1/logs](http://vendor1.com/v1/logs) \-\> attacker.url/v1/logs   
   2. Agent applies remote config, trusting the OpAMP Server as the trusted root, and start sending telemetry to an entirely different endpoint (which could include sensitive/customer data if the user is not properly redacting their telemetry)  
3. Unintended binary update (partially mitigated by DownloadableFile.signature)  
   1. Server sends PackagesAvailable with an unexpected distribution of OTel Collector  
   2. If the URL/database where the OpAMP agent matches the expected signature to a given binary for a distribution is compromised, the DownloadableFile.signature could appear to be valid, so the OpAMP Agent would download and install this binary

Each of these issues stems from the OpAMP Server being trusted as both the distribution server for OpAMP messages, as well as the trusted root authority. 

### Mitigation/Solution

Beyond just the signature field for a DownloadableFile message, there should be a way to separate the concerns of the trusted root authority from the distribution server, and for an OpAMP Agent to determine that any message has in fact originated from this trusted root. This might include using a pre-coordinated root certificate between OpAMP Agent and OpAMP Server, or exchanging this certificate upon initial connection between the Agent and Server, but the exact implementation details of any solution are outside the scope of this issue.

If an OpAMP Server operator (or vendor) chooses to adopt the model of a separate trusted root authority for messages, they could separate concerns by signing messages to send to a particular OpAMP Agent in a separate environment from the OpAMP server itself.   
Additionally, a message signing service/process could enforce schemas or field validation (e.g. “endpoint” config values must have [securecorp.com](http://securecorp.com) as base url), and refuse to sign messages that don’t match this requirement. In this manner, an attacker gaining access to the OpAMP Server would not be able to sign messages that do not meet strict validation/type requirements.   
If the OpAMP Agent has a method to validate the trusted root certificate, and that certificate is not present or is invalid, the Agent would reject this message and could immediately terminate the connection, preventing further attacks.

### Current Model vs. Proposed Model

The diagram below shows the way that this model can reduce the risk of an unauthorized actor sending unintended commands to an OpAMP Agent.  

Image

liustanley:
Resolves https://github.com/open-telemetry/opamp-spec/issues/265, see [proposal document](https://docs.google.com/document/d/10IwvaEMs-CqE6OmTcGJHMf7TwJJPqvAabCjMQ5OTcP4/edit?tab=t.8gtjif5uhm5o) for context.

## Summary
This PR introduces an optional, end-to-end integrity mechanism for `ServerToAgent` messages based on X.509 certificate chains. When both the Server and the Agent opt in, every `ServerToAgent` message sent after the initial handshake carries a signature that the Agent verifies against a pre-configured trust anchor (a CA certificate) that is **distinct** from the TLS certificate authority used to establish the transport.

The mechanism allows OpAMP deployments to separate the _distribution server_ from the _authoritative source of OpAMP messages_. A compromised distribution server cannot, in this model, push arbitrary configuration or commands to an Agent, because every message must be signed by a key whose trust chain validates against the Agent's pre-configured root.

Strict opt-in: implementations that do not implement Message Attestation require no changes. Existing OpAMP deployments are unaffected until both sides opt in.

## Changes
Wire-level changes (all Status: [Development]):
* AgentCapabilities.RequiresPayloadTrustVerification = 0x00010000
* ServerCapabilities.OffersPayloadTrustVerification = 0x00000080
* ServerToAgent.trust_chain_response = 12 (new TrustChainResponse message)
* ServerToAgent.signature = 13

The new "Message Attestation" section in `specification.md` covers the threat model, trust model, opt-in semantics, capability negotiation, connection-time handshake, in-session per-message verification, the heartbeat response exemption (an empty heartbeat response — only `instance_uid` set — MAY be sent unsigned; anything else unsigned is rejected), algorithm selection (any X.509-supported), certificate requirements (including multi-SAN signing certificates so a single key can serve an OpAMP gateway/proxy in front of the origin server without relaxing the Agent's hostname check), failure modes (all result in agent disconnect), and out-of-scope items.

### Client-facing changes
To enable verification of messages, the OpAMP client is configured with a root certificate to trust at startup via a config value:
```
extensions:
  opamp:
    server:
      payload_ca: /etc/opamp/securecorp-ca.pem
      ws:
        endpoint: wss://127.0.0.1:4320/v1/opamp
```
When provided, the client will indicate to the server that it requires authenticated payloads, and use the provided root certificate to verify messages the server sends.

## Motivation
The OpAMP Server is currently treated as both a point of distribution and the authoritative source of OpAMP messages. If an OpAMP Server is compromised, clients can be subject to significant security vulnerabilities such as forced restarts, breaches of sensitive information, and unintended binary updates. By requiring messages to be signed by an authoritative source, an attacker gaining access to the OpAMP Server would not be able to sign messages that do not meet strict validation/type requirements. OpAMP Agents will also have the capability to reject messages with invalid certificates, and can immediately terminate the connection, preventing further attacks.

### DIFF
--- a/proto/opamp.proto
+++ b/proto/opamp.proto
@@ -351,17 +351,15 @@ message SignedServerToAgent {
     // and then unmarshals them into a ServerToAgent for normal processing.
     bytes payload = 1;
 
-    // Detached signature over the bytes of the payload field. MAY be empty
-    // on the first SignedServerToAgent of a connection: trust on the first
-    // message is established by validating the certificate chain carried in
-    // trust_chain_response against the Agent's pre-configured payload trust
-    // anchor. MUST be present and verifiable on every subsequent
-    // SignedServerToAgent.
+    // Detached signature over the bytes of the payload field. MUST be
+    // present and verifiable on every SignedServerToAgent, including the
+    // first. There is no exception for the first message.
     bytes signature = 2;
 
     // Sent only in the first SignedServerToAgent on a connection. Carries
-    // the signing certificate chain the Agent will use to verify signatures
-    // on subsequent messages. If the Agent set
+    // the signing certificate chain the Agent uses to validate the leaf
+    // certificate and verify signatures on all messages including this one.
+    // If the Agent set
     // RequiresPayloadTrustVerification but the first SignedServerToAgent
     // does not include a usable trust_chain_response, the Agent MUST
     // terminate the connection.
--- a/proto/opamp.proto
+++ b/proto/opamp.proto
@@ -329,6 +329,19 @@ message TrustChainResponse {
     // satisfy the trust chain request. If error_message is set, the Agent
     // MUST terminate the connection.
     string error_message = 2;
+
+    // PEM-encoded root CA certificate used as the payload trust anchor.
+    // Set only during Trust On First Use (TOFU) enrollment: the Server
+    // includes this when the Agent has advertised
+    // AgentCapabilities_AcceptsPayloadTrustAnchorTOFU and the Agent has no
+    // pre-configured trust anchor. The Agent MUST persist this certificate
+    // and use it as the payload trust anchor for all subsequent connections.
+    // The Agent MUST NOT update a previously persisted or operator-configured
+    // trust anchor — if the Agent already has a trust anchor this field MUST
+    // be ignored. To rotate a TOFU-enrolled anchor the operator deletes the
+    // persisted anchor file and restarts the Agent (out-of-band rotation).
+    // See the Trust On First Use (TOFU) section of the specification.
+    bytes tofu_trust_anchor = 3;
 }
 
 // SignedServerToAgent wraps a ServerToAgent message when the payload trust
@@ -860,6 +873,15 @@ enum AgentCapabilities {
     // the connection. See the Message Attestation section of the specification.
     // Status: [Development]
     AgentCapabilities_RequiresPayloadTrustVerification = 0x00010000;
+    // The Agent supports Trust On First Use (TOFU) enrollment for the payload
+    // trust anchor. When set alongside
+    // AgentCapabilities_RequiresPayloadTrustVerification it signals that the
+    // Agent has no pre-configured trust anchor and asks the Server to include
+    // the root CA in trust_chain_response.tofu_trust_anchor so the Agent can
+    // bootstrap and persist it. MUST NOT be set if the Agent already has a
+    // persisted or operator-configured trust anchor.
+    // Status: [Development]
+    AgentCapabilities_AcceptsPayloadTrustAnchorTOFU = 0x00020000;
     // Add new capabilities here, continuing with the least significant unused bit.
 }
 
--- a/proto/opamp.proto
+++ b/proto/opamp.proto
@@ -318,16 +318,12 @@ enum ServerCapabilities {
 // specification.
 // Status: [Development]
 message TrustChainResponse {
-    message Certificate {
-        // The certificate in DER format.
-        bytes der_data = 1;
-    }
-
-    // The certificate chain, ordered from the first intermediate certificate
-    // down to the signing leaf certificate. The root certificate is excluded;
-    // the Agent already possesses the root as its pre-configured payload
-    // trust anchor.
-    repeated Certificate certificate_chain = 1;
+    // PEM-encoded certificate chain, ordered from the first intermediate
+    // certificate down to the signing leaf certificate. The root certificate
+    // is excluded; the Agent already possesses the root as its pre-configured
+    // payload trust anchor. Multiple certificates are concatenated in a single
+    // PEM blob, consistent with the encoding used by TLSCertificate.
+    bytes certificate_chain = 1;
 
     // Human-readable error message indicating why the Server could not
     // satisfy the trust chain request. If error_message is set, the Agent
--- a/proto/opamp.proto
+++ b/proto/opamp.proto
@@ -243,6 +243,23 @@ message ServerToAgent {
     // A custom message sent from the Server to an Agent.
     // Status: [Development]
     CustomMessage custom_message = 11;
+
+    // Sent by the Server in its first ServerToAgent message in response to an
+    // Agent that has set the RequiresPayloadTrustVerification capability.
+    // Carries the signing certificate chain the Agent will use to verify
+    // subsequent ServerToAgent messages. If the Agent set
+    // RequiresPayloadTrustVerification but the first ServerToAgent does not
+    // include trust_chain_response, the Agent MUST terminate the connection.
+    // See the Message Attestation section of the specification.
+    // Status: [Development]
+    TrustChainResponse trust_chain_response = 12;
+
+    // The signature of this ServerToAgent message. The exact bytes that are
+    // signed and the verification procedure are defined in the
+    // Message Attestation section of the specification. Set only after the
+    // payload trust verification handshake has completed successfully.
+    // Status: [Development]
+    bytes signature = 13;
 }
 
 enum ServerToAgentFlags {
@@ -289,8 +306,36 @@ enum ServerCapabilities {
     // The Server can accept ConnectionSettingsRequest and respond with an offer.
     // Status: [Development]
     ServerCapabilities_AcceptsConnectionSettingsRequest = 0x00000040;
+    // The Server can respond to the payload trust verification handshake and
+    // sign every ServerToAgent message it sends after the handshake. See the
+    // Message Attestation section of the specification.
+    // Status: [Development]
+    ServerCapabilities_OffersPayloadTrustVerification = 0x00000080;
 
     // Add new capabilities here, continuing with the least significant unused bit.
+}
+
+// TrustChainResponse carries the signing certificate chain used by the Server
+// to sign subsequent ServerToAgent messages, as part of the payload trust
+// verification handshake. See the Message Attestation section of the
+// specification.
+// Status: [Development]
+message TrustChainResponse {
+    message Certificate {
+        // The certificate in DER format.
+        bytes der_data = 1;
+    }
+
+    // The certificate chain, ordered from the first intermediate certificate
+    // down to the signing leaf certificate. The root certificate is excluded;
+    // the Agent already possesses the root as its pre-configured payload
+    // trust anchor.
+    repeated Certificate certificate_chain = 1;
+
+    // Human-readable error message indicating why the Server could not
+    // satisfy the trust chain request. If error_message is set, the Agent
+    // MUST terminate the connection.
+    string error_message = 2;
 }
 
 // The OpAMPConnectionSettings message is a collection of fields which comprise an
@@ -781,6 +826,12 @@ enum AgentCapabilities {
     // The agent will report ConnectionSettingsOffers status via AgentToServer.connection_settings_status field.
     // Status: [Development]
     AgentCapabilities_ReportsConnectionSettingsStatus = 0x00008000;
+    // The Agent requires the payload trust verification handshake on connection
+    // and signature verification on every subsequent ServerToAgent message.
+    // If the Server does not offer this capability, the Agent MUST terminate
+    // the connection. See the Message Attestation section of the specification.
+    // Status: [Development]
+    AgentCapabilities_RequiresPayloadTrustVerification = 0x00010000;
     // Add new capabilities here, continuing with the least significant unused bit.
 }
 
--- a/proto/opamp.proto
+++ b/proto/opamp.proto
@@ -244,22 +244,14 @@ message ServerToAgent {
     // Status: [Development]
     CustomMessage custom_message = 11;
 
-    // Sent by the Server in its first ServerToAgent message in response to an
-    // Agent that has set the RequiresPayloadTrustVerification capability.
-    // Carries the signing certificate chain the Agent will use to verify
-    // subsequent ServerToAgent messages. If the Agent set
-    // RequiresPayloadTrustVerification but the first ServerToAgent does not
-    // include trust_chain_response, the Agent MUST terminate the connection.
-    // See the Message Attestation section of the specification.
-    // Status: [Development]
-    TrustChainResponse trust_chain_response = 12;
-
-    // The signature of this ServerToAgent message. The exact bytes that are
-    // signed and the verification procedure are defined in the
-    // Message Attestation section of the specification. Set only after the
-    // payload trust verification handshake has completed successfully.
-    // Status: [Development]
-    bytes signature = 13;
+    // Field numbers 12 and 13 were briefly assigned to inline payload
+    // trust verification metadata (trust_chain_response and signature)
+    // in an earlier draft of the Message Attestation spec. The design
+    // moved that metadata onto a separate envelope message
+    // (SignedServerToAgent) to enable detached signing. The field
+    // numbers are reserved here to prevent accidental reuse.
+    reserved 12, 13;
+    reserved "trust_chain_response", "signature";
 }
 
 enum ServerToAgentFlags {
@@ -336,6 +328,43 @@ message TrustChainResponse {
     // satisfy the trust chain request. If error_message is set, the Agent
     // MUST terminate the connection.
     string error_message = 2;
+}
+
+// SignedServerToAgent wraps a ServerToAgent message when the payload trust
+// verification handshake has been negotiated between Server and Agent. When
+// both AgentCapabilities_RequiresPayloadTrustVerification (set by the Agent)
+// and ServerCapabilities_OffersPayloadTrustVerification (set by the Server)
+// are advertised, every Server-to-Agent message on the connection is wrapped
+// in SignedServerToAgent.
+//
+// The signature is computed and verified over the bytes of the payload field
+// exactly as they appear on the wire (a "detached" signature). This avoids
+// any dependency on canonical protobuf encoding, which is not guaranteed
+// across protobuf library versions, schema changes, or build flags.
+//
+// See the Message Attestation section of the specification.
+// Status: [Development]
+message SignedServerToAgent {
+    // Serialised bytes of a ServerToAgent message. The Agent verifies the
+    // detached signature over these exact bytes, without re-marshalling,
+    // and then unmarshals them into a ServerToAgent for normal processing.
+    bytes payload = 1;
+
+    // Detached signature over the bytes of the payload field. MAY be empty
+    // on the first SignedServerToAgent of a connection: trust on the first
+    // message is established by validating the certificate chain carried in
+    // trust_chain_response against the Agent's pre-configured payload trust
+    // anchor. MUST be present and verifiable on every subsequent
+    // SignedServerToAgent.
+    bytes signature = 2;
+
+    // Sent only in the first SignedServerToAgent on a connection. Carries
+    // the signing certificate chain the Agent will use to verify signatures
+    // on subsequent messages. If the Agent set
+    // RequiresPayloadTrustVerification but the first SignedServerToAgent
+    // does not include a usable trust_chain_response, the Agent MUST
+    // terminate the connection.
+    TrustChainResponse trust_chain_response = 3;
 }
 
 // The OpAMPConnectionSettings message is a collection of fields which comprise an
--- a/proto/opamp.proto
+++ b/proto/opamp.proto
@@ -249,10 +249,14 @@ message ServerToAgent {
     // trust verification metadata (trust_chain_response and signature)
     // in an earlier draft of the Message Attestation spec. The design
     // moved that metadata onto a separate envelope message
-    // (SignedServerToAgent) to enable detached signing. The field
-    // numbers are reserved here to prevent accidental reuse.
-    reserved 12, 13;
-    reserved "trust_chain_response", "signature";
+    // (SignedServerToAgent) to enable detached signing. Field numbers
+    // 14–16 are assigned to SignedServerToAgent.payload, .signature,
+    // and .trust_chain_response respectively, so that the two messages
+    // have non-overlapping field numbers and cannot be accidentally
+    // misinterpreted by a parser expecting the other type. All six
+    // field numbers are reserved here to prevent reuse on ServerToAgent.
+    reserved 12, 13, 14, 15, 16;
+    reserved "trust_chain_response", "signature", "payload";
 }
 
 enum ServerToAgentFlags {
@@ -349,12 +353,12 @@ message SignedServerToAgent {
     // Serialised bytes of a ServerToAgent message. The Agent verifies the
     // detached signature over these exact bytes, without re-marshalling,
     // and then unmarshals them into a ServerToAgent for normal processing.
-    bytes payload = 1;
+    bytes payload = 14;
 
     // Detached signature over the bytes of the payload field. MUST be
     // present and verifiable on every SignedServerToAgent, including the
     // first. There is no exception for the first message.
-    bytes signature = 2;
+    bytes signature = 15;
 
     // Sent only in the first SignedServerToAgent on a connection. Carries
     // the signing certificate chain the Agent uses to validate the leaf
@@ -363,7 +367,7 @@ message SignedServerToAgent {
     // RequiresPayloadTrustVerification but the first SignedServerToAgent
     // does not include a usable trust_chain_response, the Agent MUST
     // terminate the connection.
-    TrustChainResponse trust_chain_response = 3;
+    TrustChainResponse trust_chain_response = 16;
 }
 
 // The OpAMPConnectionSettings message is a collection of fields which comprise an
--- a/proto/opamp.proto
+++ b/proto/opamp.proto
@@ -369,10 +369,18 @@ message SignedServerToAgent {
     // first. There is no exception for the first message.
     bytes signature = 15;
 
-    // Sent only in the first SignedServerToAgent on a connection. Carries
-    // the signing certificate chain the Agent uses to validate the leaf
-    // certificate and verify signatures on all messages including this one.
-    // If the Agent set
+    // Sent in the first SignedServerToAgent on a connection, and again
+    // whenever the signing certificate chain changes (signing certificate
+    // rotation). Carries the signing certificate chain the Agent uses to
+    // validate the leaf certificate and verify signatures. The root
+    // (payload trust anchor) never changes; only the chain to it does.
+    // When trust_chain_response carries a chain that differs from the one
+    // the Agent currently has pinned, the Agent MUST validate it against
+    // its pre-configured trust anchor and re-pin the resulting leaf; a
+    // chain identical to the pinned one is a no-op. This lets a
+    // mid-connection rotation avoid dropping the connection. The HTTP
+    // transport MAY re-send the chain on every response, so the Agent MUST
+    // treat an unchanged chain as a no-op. If the Agent set
     // RequiresPayloadTrustVerification but the first SignedServerToAgent
     // does not include a usable trust_chain_response, the Agent MUST
     // terminate the connection.
--- a/proto/opamp.proto
+++ b/proto/opamp.proto
@@ -245,18 +245,13 @@ message ServerToAgent {
     // Status: [Development]
     CustomMessage custom_message = 11;
 
-    // Field numbers 12 and 13 were briefly assigned to inline payload
-    // trust verification metadata (trust_chain_response and signature)
-    // in an earlier draft of the Message Attestation spec. The design
-    // moved that metadata onto a separate envelope message
-    // (SignedServerToAgent) to enable detached signing. Field numbers
-    // 14–16 are assigned to SignedServerToAgent.payload, .signature,
-    // and .trust_chain_response respectively, so that the two messages
-    // have non-overlapping field numbers and cannot be accidentally
-    // misinterpreted by a parser expecting the other type. All six
-    // field numbers are reserved here to prevent reuse on ServerToAgent.
-    reserved 12, 13, 14, 15, 16;
-    reserved "trust_chain_response", "signature", "payload";
+    // Field numbers 14, 15, and 16 are reserved on ServerToAgent because
+    // they are used by SignedServerToAgent, the envelope message that
+    // carries a serialised ServerToAgent in its payload field. Keeping the
+    // field numbers of the two messages non-overlapping ensures that if the
+    // bytes of one are ever decoded as the other, the mismatched fields are
+    // seen as unknown fields and ignored rather than silently misinterpreted.
+    reserved 14, 15, 16;
 }
 
 enum ServerToAgentFlags {
--- a/specification.md
+++ b/specification.md
@@ -3571,9 +3571,6 @@ Agent by default. The capabilities should be opt-in by the user.
 
 Any executable code that is part of a package should be signed
 to prevent a compromised Server from delivering malicious code to the Agent.
-For end-to-end integrity of OpAMP messages themselves (including remote
-configuration offers, package offers, and Server-issued commands), see the
-[Message Attestation](#message-attestation) section.
 We recommend the following:
 
 * Any downloadable executable code (e.g. executable packages)
@@ -3592,6 +3589,10 @@ We recommend the following:
   code that it runs as external processes) at the minimum possible privilege to
   prevent the code from accessing sensitive files or perform high privilege
   operations. The Agent should not run downloaded code as root user.
+
+For end-to-end integrity of OpAMP messages themselves (including remote
+configuration offers, package offers, and Server-issued commands), see the
+[Message Attestation](#message-attestation) section.
 
 ## Message Attestation
 
@@ -3707,10 +3708,11 @@ sending plain `ServerToAgent` messages, and the Agent keeps parsing
 them as such. Implementations that do not implement Message Attestation
 therefore see no wire-format change at all.
 
-A Server may advertise `OffersPayloadTrustVerification` and only wrap
-its outbound messages in `SignedServerToAgent` on connections from
-Agents that have set `RequiresPayloadTrustVerification` — there is no
-per-message cost for advertising the capability to non-opted-in Agents.
+A Server SHOULD only advertise `OffersPayloadTrustVerification` to
+Agents that have set `RequiresPayloadTrustVerification`, as advertising
+to non-opted-in Agents adds unnecessary bits to the capabilities field.
+A Server MUST only wrap its outbound messages in `SignedServerToAgent`
+on connections from Agents that have set `RequiresPayloadTrustVerification`.
 
 ### Capability Negotiation
 
@@ -3718,7 +3720,7 @@ per-message cost for advertising the capability to non-opted-in Agents.
 | --- | --- | --- |
 | No | No | Plain OpAMP. The Server sends `ServerToAgent` messages on the wire. Today's behaviour. |
 | No | Yes | Plain OpAMP. The Server is capable of signing but the Agent has not opted in, so the Server sends `ServerToAgent` messages on the wire (not `SignedServerToAgent`). |
-| Yes | No | The Server does not send `SignedServerToAgent`. The Agent receives a plain `ServerToAgent` where it expected a `SignedServerToAgent` envelope, and MUST terminate the connection. |
+| Yes | No | The Server does not send `SignedServerToAgent`. The Agent receives a plain `ServerToAgent` where it expected a `SignedServerToAgent` envelope. The Agent MUST treat this as a failure as described in [Failure Modes](#failure-modes). |
 | Yes | Yes | Every Server-to-Agent message on the connection is wrapped in `SignedServerToAgent`. Handshake on the first message (carries `trust_chain_response`); per-message detached signatures thereafter. Specified in the remainder of this section. |
 
 ### Connection-Time Handshake
@@ -3772,8 +3774,12 @@ first such envelope carries the signing certificate chain in
      `id-kp-codeSigning` EKU, failed name/policy constraints, or
      revocation — the Agent MUST terminate the connection.
    * On successful validation, the Agent stores the validated leaf
-     certificate (or its public key) for the duration of the
-     connection.
+     certificate (or its public key) for the duration of the session.
+     For WebSocket transport the session ends when the connection closes.
+     For HTTP transport the session persists across polling requests and
+     ends when the Agent restarts or explicitly reconnects. On
+     reconnection, the handshake is repeated and the Server presents a
+     fresh certificate chain.
    * The Agent MUST verify `signature` over the `payload` bytes using
      the public key of the validated leaf certificate. If `signature`
      is absent or verification fails, the Agent MUST terminate the
@@ -3786,13 +3792,13 @@ first such envelope carries the signing certificate chain in
 ```protobuf
 message SignedServerToAgent {
     // Serialised bytes of a ServerToAgent message.
-    bytes payload = 1;
+    bytes payload = 14;
 
     // Detached signature over the bytes of the payload field.
-    bytes signature = 2;
+    bytes signature = 15;
 
     // Sent only in the first SignedServerToAgent on a connection.
-    TrustChainResponse trust_chain_response = 3;
+    TrustChainResponse trust_chain_response = 16;
 }
 ```
 
@@ -3934,9 +3940,11 @@ they evolve.
 The signing leaf certificate MUST:
 
 * Include the Extended Key Usage extension with `id-kp-codeSigning`
-  (`1.3.6.1.5.5.7.3.3`) in its list of allowed usages. This prevents a
-  certificate intended for TLS server authentication from being used
-  to sign OpAMP messages.
+  (`1.3.6.1.5.5.7.3.3`) in its list of allowed usages. TLS server
+  certificates carry `id-kp-serverAuth` instead; because the Agent
+  requires `id-kp-codeSigning` during chain validation, a TLS
+  certificate will fail the EKU check and cannot be repurposed as a
+  signing certificate.
 * Be within its validity window
   (`notBefore <= currentTime < notAfter`) at every verification.
 
@@ -3952,13 +3960,14 @@ short-lived certificates as an alternative to active revocation.
 
 ### Failure Modes
 
-Every failure listed below MUST cause the Agent to terminate the OpAMP
-connection. The Agent SHOULD reconnect using exponential backoff consistent
-with normal OpAMP reconnection behaviour; on reconnection the Server presents
-a (potentially rotated) chain on the new handshake. Because these failures
-are detected locally by the Agent, no dedicated error-response field is
-required — the Agent terminates the connection without waiting for a
-`ServerToAgent.error_response`.
+Every failure listed below MUST cause the Agent to terminate the current
+exchange without processing the received payload. For WebSocket transport,
+the Agent closes the WebSocket connection. For HTTP transport, the Agent
+discards the response body. In both cases, the Agent SHOULD reconnect using
+exponential backoff consistent with normal OpAMP reconnection behaviour; on
+reconnection the Server presents a (potentially rotated) chain on the new
+handshake. Because these failures are detected locally by the Agent, no
+dedicated error-response field is required.
 
 | Failure | When detected |
 | --- | --- |
--- a/specification.md
+++ b/specification.md
@@ -3665,7 +3665,7 @@ The Server is independently configured with a signing key and its
 corresponding certificate chain that validates back to the payload trust
 anchor.
 
-The payload trust anchor is NEVER pushed by the Server. In particular,
+The payload trust anchor is NEVER sent from the Server to the Agent. In particular,
 no field of any `ServerToAgent` message — including
 [`OpAMPConnectionSettings`](#opampconnectionsettings) — may be used to
 update or replace the Agent's payload trust anchor. This is a deliberate
@@ -3764,12 +3764,15 @@ first such envelope carries the signing certificate chain in
      `SignedServerToAgent`.)
    * If `trust_chain_response.error_message` is non-empty, the Agent
      MUST terminate the connection.
-   * Otherwise the Agent performs standard X.509 path validation of
-     `certificate_chain` using the pre-configured payload trust anchor
-     as the trust root. If validation fails — for any reason, including
-     expired leaf, expired intermediate, unknown issuer, missing
-     `id-kp-codeSigning` Extended Key Usage on the leaf, or revocation
-     — the Agent MUST terminate the connection.
+   * Otherwise the Agent MUST perform X.509 certification path
+     validation as defined in [RFC 5280 §6](https://datatracker.ietf.org/doc/html/rfc5280#section-6),
+     using the pre-configured payload trust anchor as the sole trust
+     anchor, `certificate_chain` as the intermediate and leaf
+     certificates, and `id-kp-codeSigning` (`1.3.6.1.5.5.7.3.3`) in the
+     acceptable EKU set. If path validation fails for any reason —
+     including expired certificate, unknown issuer, missing
+     `id-kp-codeSigning` EKU, failed name/policy constraints, or
+     revocation — the Agent MUST terminate the connection.
    * On successful validation, the Agent stores the validated leaf
      certificate (or its public key) for the duration of the
      connection and uses it to verify every subsequent
--- a/specification.md
+++ b/specification.md
@@ -3743,10 +3743,9 @@ first such envelope carries the signing certificate chain in
      satisfy the trust chain request (for example, because its signing
      key is unavailable). When `error_message` is non-empty,
      `certificate_chain` SHOULD be empty.
-   * MAY leave `signature` empty on this first envelope. Trust on the
-     first message is established by chain validation against the
-     pre-configured trust anchor, not by a self-signature. Every
-     subsequent `SignedServerToAgent` MUST carry a valid signature.
+   * MUST set `signature` to a valid detached signature over the
+     `payload` bytes on this first envelope and on every subsequent
+     envelope. There is no exception for the first message.
    * Sets `payload` to the marshalled bytes of a `ServerToAgent`
      message containing whatever the Server would have sent had
      signing not been negotiated (for example, an empty `ServerToAgent`
@@ -3774,8 +3773,11 @@ first such envelope carries the signing certificate chain in
      revocation — the Agent MUST terminate the connection.
    * On successful validation, the Agent stores the validated leaf
      certificate (or its public key) for the duration of the
-     connection and uses it to verify every subsequent
-     `SignedServerToAgent`.
+     connection.
+   * The Agent MUST verify `signature` over the `payload` bytes using
+     the public key of the validated leaf certificate. If `signature`
+     is absent or verification fails, the Agent MUST terminate the
+     connection.
    * The Agent then unmarshals the `payload` bytes into a
      `ServerToAgent` and processes it normally.
 
@@ -3804,17 +3806,18 @@ these bytes into a `ServerToAgent` for normal processing.
 
 ##### SignedServerToAgent.signature
 
-Detached signature over the bytes of `payload`. MAY be empty on the
-first `SignedServerToAgent` of a connection — chain validation against
-the pre-configured trust anchor establishes initial trust. MUST be
-present and verifiable on every subsequent message.
+Detached signature over the bytes of `payload`. MUST be present and
+verifiable on every `SignedServerToAgent`, including the first. There
+is no exception for the first message — the Server signs the payload
+immediately using the signing key whose certificate chain is carried in
+`trust_chain_response`.
 
 ##### SignedServerToAgent.trust_chain_response
 
 Sent only in the first `SignedServerToAgent` on a connection. Carries
-the signing certificate chain the Agent will use to verify signatures
-on subsequent messages. See
-[TrustChainResponse Message](#trustchainresponse-message).
+the signing certificate chain the Agent uses to validate the leaf
+certificate and verify signatures on all messages including this one.
+See [TrustChainResponse Message](#trustchainresponse-message).
 
 #### TrustChainResponse Message
 
@@ -3887,8 +3890,7 @@ The Server produces a `SignedServerToAgent` as follows:
    from step 3. On the first message, also set `trust_chain_response`.
 5. Marshal and send the `SignedServerToAgent`.
 
-The Agent verifies a received `SignedServerToAgent` (after the first)
-as follows:
+The Agent verifies a received `SignedServerToAgent` as follows:
 
 1. Parse the wire bytes as a `SignedServerToAgent`. Retain the
    `payload` field's raw bytes — these are the bytes the signature
@@ -3964,8 +3966,8 @@ required — the Agent terminates the connection without waiting for a
 | First `SignedServerToAgent` does not include `trust_chain_response`. | First message. |
 | `trust_chain_response.error_message` is non-empty. | First message. |
 | Certificate chain fails X.509 path validation (expired certificate, unknown issuer, missing `id-kp-codeSigning` EKU, revoked certificate, etc.). | First message. |
-| In-session `SignedServerToAgent` lacks `signature`. | Any subsequent message. |
-| In-session `signature` does not verify against the stored leaf certificate over the received `payload` bytes. | Any subsequent message. |
+| `SignedServerToAgent` lacks `signature`. | Every message. |
+| `signature` does not verify against the stored leaf certificate over the received `payload` bytes. | Every message. |
 | Stored leaf certificate's validity window has expired since the handshake. | The next verification after expiry. |
 
 ### Out of Scope
--- a/specification.md
+++ b/specification.md
@@ -54,8 +54,6 @@ Status: [Beta]
       - [ServerToAgent.command](#servertoagentcommand)
       - [ServerToAgent.custom_capabilities](#servertoagentcustom_capabilities)
       - [ServerToAgent.custom_message](#servertoagentcustom_message)
-      - [ServerToAgent.trust_chain_response](#servertoagenttrust_chain_response)
-      - [ServerToAgent.signature](#servertoagentsignature)
     + [ServerErrorResponse Message](#servererrorresponse-message)
       - [ServerErrorResponse.type](#servererrorresponsetype)
       - [ServerErrorResponse.error_message](#servererrorresponseerror_message)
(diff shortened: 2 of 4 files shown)


specification.md
```diff
@@ -3561,6 +3617,311 @@ recommend the following:
...
+## Message Attestation
+
+**Status: [Development]**
+
+This section specifies an optional, end-to-end integrity mechanism for
+`ServerToAgent` messages based on X.509 certificate chains. When both the
+Server and the Agent opt in, every `ServerToAgent` message sent after the
+initial handshake carries a signature that the Agent verifies against a
+pre-configured trust anchor (a CA certificate) that is **distinct** from
+the TLS certificate authority used to establish the transport.
```
michel-laterman:
``` suggestion
pre-configured trust anchor (a CA certificate) that MUST differ from
the TLS certificate authority used to establish the transport.
```
you use the MUST wording later on

specification.md
```diff
@@ -3561,6 +3617,311 @@ recommend the following:
...
+The Server is independently configured with a signing key and its
+corresponding certificate chain that validates back to the payload trust
+anchor.
+
+The payload trust anchor is NEVER pushed by the Server. In particular,
+no field of any `ServerToAgent` message — including
+[`OpAMPConnectionSettings`](#opampconnectionsettings) — may be used to
+update or replace the Agent's payload trust anchor. This is a deliberate
+constraint to prevent a compromised Server from rotating the Agent onto
+an attacker-controlled trust anchor.
```
michel-laterman: What's the guidance if the trust root is set to expire?
Is some external process needed in order to update?

specification.md
```diff
@@ -3561,6 +3617,311 @@ recommend the following:
...
+   * MAY omit the [`signature`](#servertoagentsignature) field on this
+     first message. Trust on the first message is established by chain
+     validation against the pre-configured trust anchor, not by a
+     self-signature. Every subsequent `ServerToAgent` MUST be signed.
+
+3. The Agent receives the Server's first `ServerToAgent` and:
+   * If `trust_chain_response` is unset, the Agent MUST terminate the
+     connection.
+   * If `trust_chain_response.error_message` is non-empty, the Agent
+     MUST terminate the connection.
```
michel-laterman: When an agent terminates, should it retry?

Do we want to use a specific error_message attribute instead of the [ServerToAgent.error_response](https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#servertoagenterror_response)?

tigrannajaryan (changes requested):
I am blocking this for now to prevent accidental merging.

This needs a thorough review.

I am also not sure we want this now, it is not aligned with anything we have on the roadmap: https://github.com/open-telemetry/opamp-spec/pull/321

If we make the call we want this in the spec, this is going to require a prototype demonstrating it (not needed for now) and we will need to decide _when_ we want this to be added (possibly after 1.0).

specification.md
domodwyer: I'm interested in drives this requirement for separate CAs? It would require a customer that deploys message attestation to maintain / utilise two distinct CAs which is a bit of a burden.

If it is to split the security domains between the TLS + attestation, then I think we get that already using the `Key Usage` constraints on the certificates themselves:

* A server using TLS requires the `serverAuth` key usage bit to be set.
* This spec requires that the `codeSigning` bit be set.

So this lets us have two distinct chain "roles" that can chain to the same root, but are isolated from each other. You could go as far as to have the `codeSigning` intermediate CA stored separately from the `serverAuth` intermediate, physically isolating them too.

specification.md
```diff
@@ -3561,6 +3617,311 @@ recommend the following:
...
+  [`ServerToAgent.command`](#servertoagentcommand),
+  [`ServerToAgent.agent_identification`](#servertoagentagent_identification),
+  or any other field, even though such messages did not originate from
+  the intended authoritative source.
+
+* An internal attacker who can access the distribution server's stored
+  state, but not the signing keys, can still alter OpAMP messages in
+  flight.
+
+Message Attestation does **not** address:
```
domodwyer, line 3641: So I think there's two other properties that this proposal doesn't currently address that we might want to consider - they're both part of a class of attacks we expect for our internal protocol to mitigate:

* **Message replay:** if an attacker obtains a valid and signed message, they can reuse that message and send it to the same agent again, or a different agent that was not the intended recipient.
* **Rollback attack:** related to the above, a previously captured message can be sent to revert any subsequent changes (like "roll back a config change").

A classic replay attack in this context would be capturing a validly signed `ServerToAgentCommand.type = Restart` message for an Agent, and then resending it every time that agent connects, causing a DoS of the Agent using the attested message.

A more subtle attack would be to use the above replay trick to obtain a legitimate remote config message that compels an Agent to upload sensitive files to the server[^rc-security] or perform some similarly risky action, and then broadcast that command to all connected agents instead of just the intended recipient.

Notably an attacker can't change the _content_ of these messages (that would invalidate the signature). They're constrained to only replaying messages previously sent by the server.

Fixes are relatively easy: 

* To prevent cross-session replay attacks, you can make use of a session nonce ("number used once") that is randomly generated by the Agent, unique to the connection and is part of the signed `ServerToAgent` payloads sent to that Agent. Upon receiving a payload, the ID is checked to match. That prevents a message destined for Agent A from being replayed to Agent B.

* To prevent intra-session replay attacks: maintain a sequence number such that an old message cannot be resent later in the session to roll back the effects of a message (e.g. cannot "undo" a "change config" request) - [much like your car probably does](https://www.pcmag.com/news/is-your-car-key-fob-vulnerable-to-this-simple-replay-attack). There's a similar mechanism already in place for [`AgentToServer` messages](https://opentelemetry.io/docs/specs/opamp/#agenttoserversequence_num) already, but for a different purpose.

Note: there is [the `instance_id` field](https://opentelemetry.io/docs/specs/opamp/#agenttoserverinstance_uid) which goes a long way to preventing cross-Agent replay attacks, but the protocol allows the server to push an "instance ID change" request ([ref](https://opentelemetry.io/docs/specs/opamp/#servertoagentagent_identification)) which could be used to switch the Agent to an instance ID that is part of a previously captured request. I'm not familiar enough with the implementation to know if this is a viable path, and if not, the existing instance ID may suffice as a session nonce, but it should be documented as a load-bearing part of the protocol security if used as such.

[^rc-security]: https://opentelemetry.io/docs/specs/opamp/#security-considerations

specification.md
```diff
@@ -3561,6 +3617,311 @@ recommend the following:
...
+operator-managed and is supplied to the Agent through
+implementation-specific configuration (for example, a file path in the
+Agent's configuration). The payload trust anchor MUST NOT be the same
+certificate as the TLS root the Agent uses to validate the transport —
+using the same root would collapse two separate trust domains into one
+and eliminate the security benefit.
+
+The Server is independently configured with a signing key and its
+corresponding certificate chain that validates back to the payload trust
+anchor.
```
domodwyer: Is this paragraph implying that the server MUST validate payloads are valid before they are sent to the client?

If so, I would consider skipping this - it's not going to help you in the case of an attack (the attacker can simply send their message directly, bypassing the check) but it imposes a cost in the happy path as every message must be validated (latency + CPU).

specification.md
```diff
@@ -3561,6 +3617,311 @@ recommend the following:
...
+These algorithms are covered by the default X.509 stacks of Go
+(`crypto/x509`), Java (`java.security`), and Python (`cryptography`).
+
+Future algorithm additions require no change to the OpAMP protocol; new
+algorithms are signalled by the certificate and supported by stacks as
+they evolve.
+
+### Certificate Requirements
+
+The signing leaf certificate MUST:
```
domodwyer, line 4073: Another constraint: the certificate must have a `dNSName` or `iPAddress` SAN entry that matches the hostname (or IP address if IPs are used) of the opamp distribution server.

This will be required by any X509 verification library - it is what is used to ensure you're using the right certificate for the server you're connected to.

specification.md
```diff
@@ -3561,6 +3617,311 @@ recommend the following:
...
+  `ServerToAgent` message.
+* **Algorithm negotiation.** The certificate's `signatureAlgorithm` is
+  authoritative. Future algorithm support is added by certificate
+  issuers and X.509 stacks, not by changes to OpAMP.
+* **Per-message-type opt-out (signing allowlist).** Mechanisms by which
+  an Agent might accept some `ServerToAgent` message types unsigned —
+  for example, to allow a third-party fleet manager to push low-risk
+  read-only telemetry settings while still requiring authoritative
+  signatures for configuration or command messages — are deferred to a
+  follow-up specification.
```
domodwyer, line 4146: I think this presents an interesting future development if justified by usage - being able to safely delegate parts of operating the fleet to 3rd parties or higher-risk infra, while retaining control over risk factors is a nice feature.

specification.md
```diff
@@ -3561,6 +3617,311 @@ recommend the following:
...
+
+1. Construct the `ServerToAgent` message with all required fields set
+   except `signature`, which is left empty (zero-length).
+2. Serialise the message using deterministic Protocol Buffers encoding
+   (for example, Go's `proto.MarshalOptions{Deterministic: true}` or
+   the equivalent in other implementations).
+3. Sign the resulting byte string with the Server's signing private
+   key, using the signature algorithm declared by the leaf
+   certificate's `signatureAlgorithm` field.
+4. Set `signature` to the resulting signature bytes.
```
domodwyer: I think this is probably going to be problematic - the protobuf encoding does not have a  _canonical_ representation. The [docs](https://protobuf.dev/programming-guides/serialization-not-canonical/) specifically state that, even in "deterministic output" mode:

> The serializer can generate different output for many reasons, including but not limited to the following variations:
>
> * The protobuf schema changes in any way.
> * The application being built changes in any way.
> * The binary is built with different flags (eg. opt vs. debug).
> * The protobuf library is updated.

Which isn't going to provide sufficient guarantees when deployed across different versions / implementations / platforms / etc.

In general signing a payload that also contains the signature is definitely tricky and kind of forces this "serialise -> sign -> reserialise" dance that then requires the serialisation be deterministic on all platforms.

In an ideal world, the signature would be part of the _metadata_ sent alongside the signed payload, rather than within the payload itself. For the HTTP transport  that is easy - the signature can be a header value. But the WebSocket transport don't provide a metadata side channel like like HTTP headers, and the protocol doesn't currently have any provision for metadata alongside the message payload.

Here's some possible solutions - they're all shifting the signature out of the payload (to make a "detached signature") but there might be other ways to do this that I've not thought about.

### Negotiated Type Switch

After the feature flag exchange has resulted in negotiating the use of signed messages, switch to using a different type instead of `ServerToAgent` if and only if message attestation is enabled:

```protobuf
message SignedServerToAgent {
    bytes payload = 1; // Serialised ServerToAgent
    bytes signature = 2;
}

// ServerToAgent unchanged
```

Clients will need to have a new check early in the message processing path, but then processing continues as usual:

```go
payload := conn.Read();

var message ServerToAgent
if usingSigning {
    message = verify_and_parse(payload);
} else {
    message = parse_server_to_agent(payload);
}

// Continues like normal
```

Where `verify_and_parse()` parses the `SignedServerToAgent` protobuf message, and validates the signature, and then returns the inner `ServerToAgent`. The rest of the message processing continues like normal, using the `ServerToAgent` as it does today.

Pros:

* Protocol is fully backwards compatible with existing clients.

Cons:

* Protocol is now not fully described by a single type for the server -> agent direction.


### Explicit Protocol Metadata

Obviously the protocol is not yet v1, so technically a breaking change could be made to support attestation directly - either by modifying the protobuf:

```protobuf
// Redefine the top-level type:
message ServerToAgent {
    ServerToAgentPayload payload = 1;
//or  bytes payload = 1;

    // metadata here
    bytes signature = 2;
}

message ServerToAgentPayload {
    // Existing fields here
}
```

This adds user friction though and probably wouldn't be well received, but it is a "one time" cost prior to stability. 

Alternatively, by changing the [WebSocket header](https://opentelemetry.io/docs/specs/opamp/#websocket-message-format) to explicitly indicate that a signature is present alongside the encoded protobuf payload:

```
┌────────────┬────────────────────────────────────────┬───────────────────┐
│header      │varint encoded unsigned 64 bit integer  │1-10 bytes         │
├────────────┼────────────────────────────────────────┼───────────────────┤
│data        │Encoded Protobuf message,               │0 or more bytes    │
│            │either AgentToServer or ServerToAgent   │                   │
├────────────┼────────────────────────────────────────┼───────────────────┤
│metadata    │serialised metadata                     │0 or more bytes    │
└────────────┴────────────────────────────────────────┴───────────────────┘
```

They're both equivalent except the former couples the metadata to protobuf.

Pro:

* Explicit "bag of metadata" separates concerns from "payload content" in the same way HTTP / gRPC does and can be reused.

Cons:

* Breaking change - probably a difficult option at this point.

domodwyer (commented):
_Disclaimer: I provided technical help for the proposal._

I agree - I think getting a working prototype nailed down would really help shake out any issues :+1:

I understand the roadmap to V1 is already being worked on so the timing isn't great, but I do think there's value in at least investigating what a more secure protocol might look like.

The primary concern with OpAMP as it proposed for V1, is that it's a security single-point-of-failure. If you compromise the distribution server (which necessarily sits in a risky network position, reachable by many machines) then it's game over: the attacker has full control of the fleet.

This is especially bad if you're a 3rd party provider like Datadog (my employer) as a single nginx exploit or similar could yield control of many thousands of machines across many distinct organisations. This kind of exposure becomes a significant business risk and isn't something that's easy to accept - it unfortunately fundamentally restricts our adoption of the open protocol we're otherwise big fans of!

We're definitely motivated to help develop any solutions 🙏

specification.md
truthbk: If a root is expired the assumption is there will be some manual or out-of-band redistribution of the new trust root certs to the clients. Baking in rotation of the root cert into the protocol would allow for attackers to hijack the entire attestation process.

specification.md
truthbk: IHMO retries (with an exponential backoff to prevent hurting the OpAMP servers) are OK; it would allow the clients to self-recover on some class of attacks, for example DNS-based, without having to do anything specific on the client-side (in our example, once the DNS issue is resolved). 

I personally don't see a need to have a dedicated message attribute, but I don't have a strong opinion on my side.

specification.md
truthbk: I kind of feel like the negotiated type-switch would create less friction, and prioritizes the notion of backward compatibility with existing clients. I feel like it's fair to say that despite this not being v1 there are enough clients out there using former implementations that this is the more realistic/reasonable path forward.

I've updated the spec in commit https://github.com/open-telemetry/opamp-spec/pull/333/commits/1f75ca1f63a0006e627bc8f23c13a9279a1232ac to account for this feedback.

tigrannajaryan:
I took a quick look at the proposal and it is not clear to me exactly what kind of attacks this proposal prevents.

If I understand the server is expected to sign all payloads so that the client can verify. This requires the server to have access to the certificate that is used for signing. If the attack that we are attempting to prevent is performed by compromising the server, what exactly is prevented here? The attacker who can tamper with the ServerToAgent messages on the server will also gain access to the signing certificate and can sign the tampered payloads.

Unless the signing process and access to the signing certificate is separated from the server that creates ServerToAgent messages, I am not sure I understand what purpose this serves. If that separation is the intent then I do not see this described in the specification. Are you aiming for having 2 different servers, one that creates ServerToAgent payloads and another server that signs the payloads? Only the second server has access to the certificate presumably?

Please describe the mechanics of how you intend this to work in more details. An architecture diagram would be very helpful.

domodwyer:
>I took a quick look at the proposal and it is not clear to me exactly what kind of attacks this proposal prevents.

So the [spec proposal doc](https://docs.google.com/document/d/10IwvaEMs-CqE6OmTcGJHMf7TwJJPqvAabCjMQ5OTcP4/edit?tab=t.8gtjif5uhm5o#heading=h.ixgh839gvcz2) outlines some "attack vectors" it helps mitigate (which are more accurately "security concerns" - they're not all attacks) but the TLDR is that there's some benefits to traceability and auditing by cryptographically tying changes to users / infra operator identities, and additional hardening of data at rest, etc, but realistically these are probably "nice to haves" for most use OpAMP users.

The real pain point today is that a compromise of the OpAMP server (or TLS terminating proxy) yields full control of the OpAMP fleet connected to it - I think this is probably the biggest concern. It is worth noting that since my previous message an [nginx bug that enables remote code execution](https://my.f5.com/manage/s/article/K000161019) in specific configurations was identified, which if successfully exploited, would give an attacker the access they'd need to take control of the fleet today. nginx is commonly deployed as a load balancer / ingress gateway in front of the actual services like an OpAMP server.


> If I understand the server is expected to sign all payloads so that the client can verify. This requires the server to have access to the certificate that is used for signing. ...

I completely agree with your general sentiment in this paragraph - this configuration would make no sense as it provides no benefit. Opting into attestation here adds nothing over using the existing TLS setup.

> If that separation is the intent then I do not see this described in the specification. Are you aiming for having 2 different servers, one that creates ServerToAgent payloads and another server that signs the payloads? Only the second server has access to the certificate presumably?

Yes, this is generally how this would work. The split helps isolate any compromise of the distribution server, which necessarily sits at a relatively high exposure position on the "edge" of the network, where it is widely reachable (by at least every member of the fleet).

I've sketched a quick diagram of an example deployment for the sake of discussion:

image


In the above:

1. The OpAMP agent connects to the reachable OpAMP server at the network / infra team's edge.
2. The OpAMP server decides what message it wants to send to the agent (omitted from this diagram is data sourcing, etc). It builds a full `ServerToAgent` payload.
3. The OpAMP server submits this `ServerToAgent` payload to the internal policy engine; this server looks at the request, and decides if it is suitable to be sent to the specific OpAMP agent.

This is the core of the additional security - it's structurally isolating the "what can be done" decision away from the edge server that is relatively easier to gain control over. The policy engine should obviously be deployed such that it is considered secure and trustworthy - limited network access, limited authorised users, additional reviews for code changes, etc.

This policy engine can decide on what is allowed to be sent to the OpAMP agent using whatever criteria is relevant to the organisation. 

That might include factors such as:

* Denying specific messages outright (e.g. no `ServerToAgentCommand` allowed).
* Enforcing specific customer opt-ins / opt-outs ("I only want to allow using feature X, not Y").
* Enforcing security constraints on messages ("the SIEM pipeline must never be disabled").
* Limiting the types and content of custom messages / plugins.
* Only configs originating from the infra management team can be pushed to OpAMP agents.
* Only the latest version of tracers can be installed (mitigating "bring your own vulnerability" type attacks)
* etc

These constraints could be added client side, but the level of flexibility necessary to express the above (and more needed in prod envs) would require a fair bit of complexity in the Agent code, shifting the burden onto OTel rather than the organisation itself. With an attestation approach, the protocol defines a way to stamp a message as "allowed" and the client knows how to validate it - all the bespoke decision logic is maintained in the backend, by the organisation itself.

6. If accepted, the policy engine submits a signature request to a secure signer (typically a tightly controlled hardware security module) to cryptographically "stamp" the message as approved.

7. The now signed `ServerToAgent` is sent back to the agent, along with the signature it can use to validate the message was authorised.

This is of course only an example, and any actual deployment may be simpler or more complex depending what the organisation has already in place, and what risk mitigation is deemed necessary. We could definitely write some of the deployment guidelines into the spec, but as-is today it doesn't really prescribe deployments - we would definitely be happy to contribute some documentation as part of this effort though! (either in the spec or as general docs).

proto/opamp.proto
```diff
@@ -289,17 +298,82 @@ enum ServerCapabilities {
...
+// in SignedServerToAgent.
+//
+// The signature is computed and verified over the bytes of the payload field
+// exactly as they appear on the wire (a "detached" signature). This avoids
+// any dependency on canonical protobuf encoding, which is not guaranteed
+// across protobuf library versions, schema changes, or build flags.
+//
+// See the Message Attestation section of the specification.
+// Status: [Development]
+message SignedServerToAgent {
```
tigrannajaryan, line 356: I suggest that `SignedServerToAgent` message uses field numbers that are not used by `ServerToAgent` and we also declare those numbers as reserved for `ServerToAgent` to ensure there is no overlap between field numbers of these 2 messages.

This way if client and server get out of sync for whatever reason (e.g. a bug), and server sends `SignedServerToAgent` while client expects `ServerToAgent` then the client will not attempt to incorrectly interpret `SignedServerToAgent` as if it is a `ServerToAgent`. The client in this situation will be guaranteed to see an empty `ServerToAgent` which is preferable over seeing some unpredictable content of `SignedServerToAgent` potentially reinterpreted as `ServerToAgent` without hitting an unmarshaling error.

specification.md
```diff
@@ -3561,6 +3593,389 @@ recommend the following:
...
+**Status: [Development]**
+
+This section specifies an optional, end-to-end integrity mechanism for
+`ServerToAgent` messages based on X.509 certificate chains. When both the
+Server and the Agent opt in, every `ServerToAgent` message sent after the
+initial handshake carries a signature that the Agent verifies against a
+pre-configured trust anchor (a CA certificate) that is **distinct** from
+the TLS certificate authority used to establish the transport.
+
+The mechanism allows OpAMP deployments to separate the _distribution
```
tigrannajaryan, line 3608: ~We need an example diagram here that shows what is meant by distribution server and a signing server and how the are expected to interact.~ Not needed anymore, we have that in supplementary-guidlines.

specification.md
```diff
@@ -3561,6 +3593,389 @@ recommend the following:
...
+succeeds. For connections where Message Attestation is not negotiated,
+the wire format is byte-identical to upstream OpAMP — the Server keeps
+sending plain `ServerToAgent` messages, and the Agent keeps parsing
+them as such. Implementations that do not implement Message Attestation
+therefore see no wire-format change at all.
+
+A Server may advertise `OffersPayloadTrustVerification` and only wrap
+its outbound messages in `SignedServerToAgent` on connections from
+Agents that have set `RequiresPayloadTrustVerification` — there is no
+per-message cost for advertising the capability to non-opted-in Agents.
```
tigrannajaryan: Strictly speaking there may be a tiny increased cost when `RequiresPayloadTrustVerification` is included while other higher bits are not present if this causes varint encoding of `capabilities` to spill to a larger number of bytes (3 instead 1 or 2).

More importantly, I don't see the point of advertising `RequiresPayloadTrustVerification` to agents which did not set  `OffersPayloadTrustVerification` in the request. What is the purpose of that?

specification.md
```diff
@@ -3561,6 +3593,389 @@ recommend the following:
...
+
+### Capability Negotiation
+
+| Agent `Requires` | Server `Offers` | Behaviour |
+| --- | --- | --- |
+| No | No | Plain OpAMP. The Server sends `ServerToAgent` messages on the wire. Today's behaviour. |
+| No | Yes | Plain OpAMP. The Server is capable of signing but the Agent has not opted in, so the Server sends `ServerToAgent` messages on the wire (not `SignedServerToAgent`). |
+| Yes | No | The Server does not send `SignedServerToAgent`. The Agent receives a plain `ServerToAgent` where it expected a `SignedServerToAgent` envelope, and MUST terminate the connection. |
```
tigrannajaryan: >MUST terminate the connection

What does "terminate connection" mean for plain http transport?

I think a better wording is to require the Agent to _ignore_ and _discard_ the message.

Whether the connection is terminated or no is secondary. In fact, the agent may want to keep trying until the it hits a server that is capable of signing. This is possible when rolling updates of servers are performed.

specification.md
```diff
@@ -3561,6 +3593,389 @@ recommend the following:
...
+   * Sends its first `SignedServerToAgent` containing
+     `trust_chain_response.certificate_chain` ordered from the first
+     intermediate down to the signing leaf certificate. The root
+     certificate (the Agent's pre-configured payload trust anchor)
+     MUST NOT be included in this chain.
+   * MAY set `trust_chain_response.error_message` if the Server cannot
+     satisfy the trust chain request (for example, because its signing
+     key is unavailable). When `error_message` is non-empty,
+     `certificate_chain` SHOULD be empty.
+   * MAY leave `signature` empty on this first envelope. Trust on the
```
tigrannajaryan: I don't think this is good idea. We may have OpAMP proxies in the future that operate on cleartext and can tamper with payloads. Requiring signatures on all envelopes prevents tampering of the first response by such intermediaries.

specification.md
```diff
@@ -3561,6 +3593,389 @@ recommend the following:
...
+     first message is established by chain validation against the
+     pre-configured trust anchor, not by a self-signature. Every
+     subsequent `SignedServerToAgent` MUST carry a valid signature.
+   * Sets `payload` to the marshalled bytes of a `ServerToAgent`
+     message containing whatever the Server would have sent had
+     signing not been negotiated (for example, an empty `ServerToAgent`
```
tigrannajaryan: Why this requirement? Why shouldn't the first payload be a fully functional one?

specification.md
```diff
@@ -3561,6 +3593,389 @@ recommend the following:
...
+     acknowledging the Agent's status report, or a `ServerToAgent`
+     carrying initial `remote_config`).
+
+3. The Agent receives the Server's first envelope and:
+   * Parses the bytes on the wire as `SignedServerToAgent`. If the
+     bytes cannot be parsed as `SignedServerToAgent`, or
+     `trust_chain_response` is unset, the Agent MUST terminate the
+     connection. (This is also how the Agent detects a Server that
+     does not support Message Attestation: such a Server would send a
+     plain `ServerToAgent`, which does not parse as
```
tigrannajaryan, line 3842: See my comment above about field numbers to prevent it does not accidentally parse as `SignedServerToAgent`

specification.md
```diff
@@ -3561,6 +3593,389 @@ recommend the following:
...
+   * If `trust_chain_response.error_message` is non-empty, the Agent
+     MUST terminate the connection.
+   * Otherwise the Agent performs standard X.509 path validation of
+     `certificate_chain` using the pre-configured payload trust anchor
+     as the trust root. If validation fails — for any reason, including
+     expired leaf, expired intermediate, unknown issuer, missing
+     `id-kp-codeSigning` Extended Key Usage on the leaf, or revocation
+     — the Agent MUST terminate the connection.
+   * On successful validation, the Agent stores the validated leaf
+     certificate (or its public key) for the duration of the
```
tigrannajaryan: What does "duration of connection" mean for plain http transport? Is it literal duration of TCP connection? Many http client implementations don't have direct means to detect the TCP connection lifetime. Furthermore this is even less clear when proxies and load balancers are involved.

specification.md
```diff
@@ -3561,6 +3593,389 @@ recommend the following:
...
+     plain `ServerToAgent`, which does not parse as
+     `SignedServerToAgent`.)
+   * If `trust_chain_response.error_message` is non-empty, the Agent
+     MUST terminate the connection.
+   * Otherwise the Agent performs standard X.509 path validation of
+     `certificate_chain` using the pre-configured payload trust anchor
+     as the trust root. If validation fails — for any reason, including
+     expired leaf, expired intermediate, unknown issuer, missing
+     `id-kp-codeSigning` Extended Key Usage on the leaf, or revocation
+     — the Agent MUST terminate the connection.
```
pierreprinetti: "Standard X.509 path validation" varies between implementations (EKU checking, name constraints, and basic-constraints enforcement on intermediates are commonly skipped or partial). Worth pinning to RFC 5280 §6 explicitly so the failure modes listed later become consequences of the normative reference rather than a parallel list that can drift.

```suggestion
   * Otherwise the Agent MUST perform X.509 certification path
     validation as defined in [RFC 5280 §6](https://datatracker.ietf.org/doc/html/rfc5280#section-6),
     using the pre-configured payload trust anchor as the sole trust
     anchor, `certificate_chain` as the intermediate and leaf
     certificates, and `id-kp-codeSigning` (`1.3.6.1.5.5.7.3.3`) in the
     acceptable EKU set. If path validation fails for any reason —
     including expired certificate, unknown issuer, missing
     `id-kp-codeSigning` EKU, failed name/policy constraints, or
     revocation — the Agent MUST terminate the connection.
```

tigrannajaryan:
@domodwyer thank you for your comment. It is very helpful and greatly clarifies the operation.

@liustanley do you mind adding a "supplementary-guidelines.md" to this PR and incorporate there the description from [here](https://github.com/open-telemetry/opamp-spec/pull/333#issuecomment-4544349021)? We usually include non-normative explanations like that in Otel specs (for example [here](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/supplementary-guidelines.md)). This is a complex feature and I think it warrants a guideline to help implementors.

We need more reviewers and more feedback on this proposal and having these guidelines in the PR rather than in a comment also makes it easier to review.

specification.md
```diff
@@ -3561,6 +3593,389 @@ recommend the following:
...
+
+Because the signature is detached, the Agent never needs to re-marshal
+the inner `ServerToAgent`. This eliminates any dependency on canonical
+serialisation between implementations.
+
+### Algorithm
+
+The signing algorithm is determined by the leaf certificate's
+`signatureAlgorithm` field. The OpAMP protocol does not negotiate
+algorithms.
```
pierreprinetti: Claude suggests the following, and calls it a nit... however addressing it might be a bit involved and especially when it comes to defining a hash algorithm table. I'll just leave it here below:

(what follows is entirely generated)

---

An X.509 leaf actually has two algorithm-related fields, and signatureAlgorithm is the wrong one to point at here.

- Certificate.signatureAlgorithm (RFC 5280 §4.1.1.2) is the algorithm the issuing CA used to sign the leaf. It says nothing about what the leaf's own key can sign with — a CA could sign an Ed25519 leaf using RSA-PSS.
- Certificate.tbsCertificate.subjectPublicKeyInfo.algorithm (RFC 5280 §4.1.2.7) identifies the leaf key's algorithm — this is what determines how the server signs OpAMP payloads and how the agent verifies them.

For ECDSA there's an additional wrinkle: SPKI carries the curve but not the hash, so the spec needs to say which hash pairs with each curve (the SHOULD list already implies P-256→SHA-256 and P-384→SHA-384, just worth making it normative). For RSA there's a similar ambiguity between PKCS#1 v1.5 and PSS — the SHOULD list pins v1.5, which is fine, but the Algorithm section should reference the SPKI key type plus the chosen scheme, not the cert's outer signatureAlgorithm.

Suggested rewording:

```suggestion
The signing algorithm is determined by the leaf certificate's public key (subjectPublicKeyInfo), paired with the hash function specified for that key type in the list below. The OpAMP protocol does not negotiate algorithms.
```

The same nit applies to the two other mentions of signatureAlgorithm (specification.md:3871 in the signing procedure and specification.md:3969 in the out-of-scope list).

pierreprinetti (commented):
This appears like a solid spec, with the clear premise that it wants to be lightweight. In my head I like to frame it that while TLS protects the message on the wire, message attestation is a powerful jump base for attribution and auditability of OpAMP changes. This is going to become extremely useful the more operations become AI-driven.

No algorithm negotiation also means that the burden to set up quantum-resistant crypto (very touchy topic for asymmetric primitives) remains orthogonal to this spec.

Just a couple comments inline. My third comment would have been to future-proof the signature block with something like:

```
message Signature {
    oneof kind {
        bytes detached = 1;
    }
}
Signature signature = 2;
```

but I understand that introducing new signatures [would be a backwards-compatible change](https://protobuf.dev/programming-guides/proto3/#wire-safe-changes) in gRPC even with the current code (thanks @domodwyer for educating me on the matter).

I hope this proposal finds its way to the main branch.

proto/opamp.proto
```diff
@@ -290,10 +299,75 @@ enum ServerCapabilities {
...
+    // detached signature over these exact bytes, without re-marshalling,
+    // and then unmarshals them into a ServerToAgent for normal processing.
+    bytes payload = 1;
+
+    // Detached signature over the bytes of the payload field. MAY be empty
+    // on the first SignedServerToAgent of a connection: trust on the first
+    // message is established by validating the certificate chain carried in
+    // trust_chain_response against the Agent's pre-configured payload trust
+    // anchor. MUST be present and verifiable on every subsequent
+    // SignedServerToAgent.
```
pierreprinetti: I would guard against an empty signature in the first SignedServerToAgent, when a remote_config payload is present (which if I understand correctly, is not ruled out in the certificate-bearing first message).

For the sake of simplicity, I would state that the signature must be present and verifiable on every SignedServerToAgent.

specification.md
```diff
@@ -3561,6 +3593,389 @@ recommend the following:
...
+     certificate (the Agent's pre-configured payload trust anchor)
+     MUST NOT be included in this chain.
+   * MAY set `trust_chain_response.error_message` if the Server cannot
+     satisfy the trust chain request (for example, because its signing
+     key is unavailable). When `error_message` is non-empty,
+     `certificate_chain` SHOULD be empty.
+   * MAY leave `signature` empty on this first envelope. Trust on the
+     first message is established by chain validation against the
+     pre-configured trust anchor, not by a self-signature. Every
+     subsequent `SignedServerToAgent` MUST carry a valid signature.
```
pierreprinetti: Again, the absence of a signature here doesn't appear to be coupled with the absence of a payload: which might mean that the first remote_config can be unsigned. I would remove `MAY leave signature empty on this first envelope.` or at the very minimum be explicit that no payload must be parsed by the Agent in absence of a signature when the capability is there

specification.md
```diff
@@ -3561,6 +3593,389 @@ recommend the following:
...
+here. The signature in `signature` covers these exact bytes; the Agent
+verifies the signature without re-marshalling, and then unmarshals
+these bytes into a `ServerToAgent` for normal processing.
+
+##### SignedServerToAgent.signature
+
+Detached signature over the bytes of `payload`. MAY be empty on the
+first `SignedServerToAgent` of a connection — chain validation against
+the pre-configured trust anchor establishes initial trust. MUST be
+present and verifiable on every subsequent message.
```
pierreprinetti: ditto

pierreprinetti (commented):
Why is the signature optional in the first SignedServerToAgent, considering that it may have a payload?

proto/opamp.proto
tigrannajaryan: >For the sake of simplicity, I would state that the signature must be present and verifiable on every SignedServerToAgent.

+1. I don't see a good reason to have an exception for the first message. It is simpler to specify it is always present.

specification.md
```diff
@@ -3541,8 +3570,11 @@ Agent by default. The capabilities should be opt-in by the user.
 ### Code Signing
 
 Any executable code that is part of a package should be signed
-to prevent a compromised Server from delivering malicious code to the Agent. We
-recommend the following:
+to prevent a compromised Server from delivering malicious code to the Agent.
+For end-to-end integrity of OpAMP messages themselves (including remote
+configuration offers, package offers, and Server-issued commands), see the
+[Message Attestation](#message-attestation) section.
```
tigrannajaryan: Please move this to the end of this section (around line 3595) to keep the flow about code signing uninterrupted.

pavankrish123:
Thanks @liustanley 

What's the "break glass" story for CA compromise?

If the signing CA (or an intermediate) is compromised, the operator needs agents to immediately stop trusting it. But agents will only accept messages signed by that very CA — creating a deadlock where the operator can't push a trust rotation via OpAMP, while the attacker can still produce valid signatures.

Should the spec define an out-of-band revocation mechanism, a pre-configured backup CA, or a time-bounded trust (e.g., short-lived signing certs that naturally expire)? Without this, I am guessing the failure mode of a CA compromise is arguably worse than not having attestation at all — we will loose fleet control and the attacker retains it.? Correct me if I am wrong

pierreprinetti:
> Thanks @liustanley
> 
> What's the "break glass" story for CA compromise?
> 
> If the signing CA (or an intermediate) is compromised, the operator needs agents to immediately stop trusting it. But agents will only accept messages signed by that very CA — creating a deadlock where the operator can't push a trust rotation via OpAMP, while the attacker can still produce valid signatures.
> 
> Should the spec define an out-of-band revocation mechanism, a pre-configured backup CA, or a time-bounded trust (e.g., short-lived signing certs that naturally expire)? Without this, I am guessing the failure mode of a CA compromise is arguably worse than not having attestation at all — we will loose fleet control and the attacker retains it.? Correct me if I am wrong

@pavankrish123

If an intermediate signing CA is compromised, standard X.509 revocation (CRL/OCSP) handles it; the spec already recommends this. If instead the root payload trust anchor itself is compromised, it must be replaced via the same out-of-band mechanism used to set it up. Certificate maintenance is indeed an additional operational burden; maybe it's worth mentioning that the operator should set up a CA cert rotation mechanism. However, the CA alone wouldn't let an attacker send messages to the Agent: for that, the control of the OpAMP server endpoint is required.

If an attacker gains control of both the signing CA and the OpAMP server endpoint, then it is indistinguishable from the legitimate operator, and the escape hatch has to be external to OpAMP. Even if there were a secondary CA, the Agent would still be fetching config from the compromised endpoint. In that scenario, the escape hatch to be same out-of-band mechanism (read: not OpAMP) that was used to set up the legitimate CA in the first place.

You seem to require the same escape hatch, with or without message attestation, in case of complete compromission. Except that with message attestation the scenario is arguably less likely to occur.

specification.md
liustanley: Good point, updated the spec to modify this requirement for separate CAs

specification.md
liustanley: Added a note to clarify this guidance

specification.md
liustanley: Agreed, added details on retries to the spec

liustanley:
> @liustanley do you mind adding a "supplementary-guidelines.md" to this PR and incorporate there the description from [here](https://github.com/open-telemetry/opamp-spec/pull/333#issuecomment-4544349021)? We usually include non-normative explanations like that in Otel specs (for example [here](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/supplementary-guidelines.md)). This is a complex feature and I think it warrants a guideline to help implementors.
> 
> We need more reviewers and more feedback on this proposal and having these guidelines in the PR rather than in a comment also makes it easier to review.

Thanks for the suggestion @tigrannajaryan, that makes sense with me. Just added details from @domodwyer's comment in a new file `supplementary-guidelines.md`.

specification.md
```diff
@@ -3561,6 +3593,404 @@ recommend the following:
...
+carry only `id-kp-codeSigning` and the signing private key is stored
+separately from the distribution server. The security boundary is the
+combination of private-key isolation and EKU constraints — not a
+requirement for a separate root CA.
+
+The Server is independently configured with a signing key and its
+corresponding certificate chain that validates back to the payload trust
+anchor.
+
+The payload trust anchor is NEVER pushed by the Server. In particular,
```
tigrannajaryan:
```suggestion
The payload trust anchor is NEVER sent from the Server to the Agent. In particular,
```

specification.md
```diff
@@ -3561,6 +3593,404 @@ recommend the following:
...
+operator-managed and is supplied to the Agent through
+implementation-specific configuration (for example, a file path in the
+Agent's configuration). The payload trust anchor MUST NOT carry the `id-kp-serverAuth`
+Extended Key Usage, ensuring it cannot double as a TLS CA certificate.
+Operators MAY root both the TLS chain and the signing chain from the
+same root CA, provided the signing intermediate and leaf certificates
+carry only `id-kp-codeSigning` and the signing private key is stored
+separately from the distribution server. The security boundary is the
+combination of private-key isolation and EKU constraints — not a
+requirement for a separate root CA.
```
tigrannajaryan, line 3664: I don't understand this piece (likely mostly due to my ignorance). Can you add some explanation in the supplementary-guidelines to help readers who don't have in-depth experience in the topic?

specification.md
```diff
@@ -3561,6 +3593,404 @@ recommend the following:
...
+no field of any `ServerToAgent` message — including
+[`OpAMPConnectionSettings`](#opampconnectionsettings) — may be used to
+update or replace the Agent's payload trust anchor. This is a deliberate
+constraint to prevent a compromised Server from rotating the Agent onto
+an attacker-controlled trust anchor.
+
+When the payload trust anchor approaches expiry or must be replaced,
+rotation is handled by the same out-of-band mechanism used to provision
+it initially (for example, updating the certificate file in the Agent's
+configuration and restarting the Agent). Operators SHOULD plan for this
```
tigrannajaryan, line 3682: Another approach is to hard-code the CA certificate in the Agent and use Agent executable updates to push new versions of the Agent (and thus new CA certificate).

specification.md
```diff
@@ -3561,6 +3593,404 @@ recommend the following:
...
+
+Marshalled bytes of an inner `ServerToAgent` message. The Server
+marshals the inner `ServerToAgent` once and places the resulting bytes
+here. The signature in `signature` covers these exact bytes; the Agent
+verifies the signature without re-marshalling, and then unmarshals
+these bytes into a `ServerToAgent` for normal processing.
+
+##### SignedServerToAgent.signature
+
+Detached signature over the bytes of `payload`. MAY be empty on the
```
tigrannajaryan: See my comment above. I don't think we want to allow this to be empty.

specification.md
```diff
@@ -3561,6 +3593,404 @@ recommend the following:
...
+they evolve.
+
+### Certificate Requirements
+
+The signing leaf certificate MUST:
+
+* Include the Extended Key Usage extension with `id-kp-codeSigning`
+  (`1.3.6.1.5.5.7.3.3`) in its list of allowed usages. This prevents a
+  certificate intended for TLS server authentication from being used
+  to sign OpAMP messages.
```
tigrannajaryan: I don't understand this part. Can you clarify how does it prevent this?

specification.md
tigrannajaryan: This is indeed interesting. We can imagine a simple fallback mechanism where the Agent upon detecting the Server does not support attestation or if signature cannot be verified, goes into "readonly" mode where it continues sending AgentToServer messages with status reports, etc, but ignores any remote config, commands, etc received from the Server.

This is also useful for situations when certificates expire. Instead of resulting in total loss of communication with the Agents this would result in readonly mode, greatly helping with understanding and troubleshooting the situation.

This may be a mode of operation that we allow in the spec and let implementations decide how strict they want to be in this failure mode.

tigrannajaryan:
@andykellr (who is the other spec maintainer) and I have discussed this proposal.

We believe it is an important missing capability of OpAMP and we would like to have it added.

We recently agreed on a roadmap for OpAMP. This capability is something we don't have on the roadmap. However, I think it is important enough to modify the roadmap and include it, provided that we have a good solution.

I reviewed the proposal and overall the suggested solution seems reasonable to me. I will post my detailed comments on the specifics a bit later.

To go ahead with this I would like to see the following:
- Create a prototype in opamp-go. We want to see Client and Server implementations. This needs to be a prototype, so you can skip the tests for now. Create an early draft PR to discuss the API and callbacks first.
- Modify the example in opamp-go to demonstrate how attestation works. Show the best-practice approach, introduce the separate out-of-process Policy server as described in supplementary-guidelines.md.
- Once the prototype and the example are in place, maintainers and approvers will review the whole package (including this PR) and provided everything is in order the capability will be added in `[Development]` status.
- Modify Collector [opampextension](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/opampextension) and Collector [Supervisor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/cmd/opampsupervisor) to support attestation. I need additional confirmation from @evan-bradley on this one, that he is able to review the PRs against these components, since he is the most active maintainer.

I expect that the prototype, example and final production version of opamp-go, opamp extension and Supervisor will be done by the authors of the proposal. This is a complicated feature. Current maintainers will not be able to provide the manpower to implement it themselves. We can commit to reviewing and long-term maintenance, provided that someone else implements it initially, at production quality, with necessary test coverage.

Note: this message is not a final approval and acceptance of the proposal. This is a commitment to review it and accept it _if it meets all requirements_, including new requirements that may come up in the future. It is possible that the proposal is rejected at some stage in the process I outlined above, although I commit to doing by best to make it successful, since I believe this capability is needed.

supplementary-guidelines.md
```diff
@@ -0,0 +1,86 @@
...
+from the *authorization* of messages. The distribution server remains at the
+network edge; the signing authority lives deeper in the infrastructure, away
+from the attack surface.
+
+## Example Deployment Architecture
+
+The following describes one example deployment. Actual deployments may be
+simpler or more complex depending on organisational requirements.
+
+![Example deployment architecture](https://github.com/user-attachments/assets/20fb3567-ff84-47f3-9e42-367953881232)
```
tigrannajaryan: Please convert this to ASCII chart.

Also mirror left-to-right. We normally put agent on the left, server on the right in other diagrams in this repo.

supplementary-guidelines.md
```diff
@@ -0,0 +1,86 @@
...
+2. The OpAMP server constructs a `ServerToAgent` payload to send to the Agent.
+3. Before sending, the OpAMP server submits the payload to an internal policy
+   engine.
+4. The policy engine evaluates whether the message is permitted for this Agent,
+   applying whatever organizational constraints are relevant.
+5. If the message is approved, the policy engine submits a signature request to
+   a secure signer (typically a hardware security module).
+6. The signer returns a signature over the payload bytes.
+7. The OpAMP server wraps the payload and signature in a `SignedServerToAgent`
+   envelope and delivers it to the Agent.
```
tigrannajaryan, line 52: Can we have a sequence diagram illustrating this flow? See examples in spec, e.g. https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#opamp-connection-setting-offer-flow

proto/opamp.proto
```diff
@@ -290,10 +299,75 @@ enum ServerCapabilities {
...
 }
 
+// TrustChainResponse carries the signing certificate chain used by the Server
+// to sign subsequent ServerToAgent messages, as part of the payload trust
+// verification handshake. See the Message Attestation section of the
+// specification.
+// Status: [Development]
+message TrustChainResponse {
+    message Certificate {
+        // The certificate in DER format.
```
tigrannajaryan: We use PEM format elsewhere: https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#tlscertificatecert

Also with PEM I think the entire chain can be in one certificate PEM, so n oneed for repeated field anymore.

specification.md
tigrannajaryan: >To prevent intra-session replay attacks: maintain a sequence number such that an old message cannot be resent later in the session to roll back the effects of a message (e.g. cannot "undo" a "change config" request) - much like your car probably does. There's a similar mechanism already in place for AgentToServer messages already, but for a different purpose.

We can look into adding sequence_num to ServerToAgent to prevent the replay attacks.

There is a a bit of nuance here that needs some thought:
- We cannot require for agent-generated and server-generated sequence_num to be in sync, they will be incremented independently, particularly when WebSocket transport is in use. Nevertheless the agent can correctly verify monotonic increase of server-side generated sequence_num. 
- Need to specify what happens on reconnects, Server restarts and also when multiple Server instances are behind load balancer.

>but the protocol allows the server to push an "instance ID change" request (ref) which could be used to switch the Agent to an instance ID that is part of a previously captured request. 

This can't happen if the attestation is used. The "instance ID change" itself is subject to attestation. We must assume if the signing server is uncompromised then "instance ID change" message is legitimate.

specification.md
tigrannajaryan: I would like to discuss this. We never want the Sever to _update_ Agent's trust anchor. I agree with that.

However, it can be very valuable to support a Trust On First Use flow. If we include the trust anchor in the certificate chain that goes from Server to Agent, then on the first run the Agent can choose to trust that anchor and use it for subsequent verification.

This greatly simplifies the initial rollout of agents since it no longer requires supplying the trust anchor certificates during installation. You can install the Agent from a general repository (e.g. released by Otel) and it will obtain the trust anchor on the first run, then correctly use it for attestation. Of course the Agent is expected to _never_ update an already obtained or preconfigured trust anchor.

Yes, there is a danger the registration can happen when the distribution server is compromised, but that's generally the drawback of all TOFU flows.

The support of TOFU flow for obtaining attestation trust anchor can be a configuration option in the Agent.

We already support TOFU flows [elsewhere](https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#trust-on-first-use).

tigrannajaryan:
@liustanley please sign EasyCLA.

specification.md
liustanley: I don't have much context on SAN entries but I can't tell whether they are required by x509 verification libraries. Looking at the Go x509 verification library, it seems like the Go library doesn't check SAN unless you explicitly pass a `DNSName` in `VerifyOptions`. It seems like a SAN entry is required for a TLS handshake but not for x509 verification, but let me know if I'm misunderstanding this.

proto/opamp.proto
liustanley: Agreed, updated the spec to require that the signature must be present and verifiable on every SignedServerToAgent, including the first message

specification.md
liustanley: I don't think there is any real benefit to advertising to agents that have not opted in, will adjust this wording.

specification.md
liustanley: Good catch, updated the wording in the Failure Modes section and added a reference to the updated wording in this table.

specification.md
liustanley: "Duration of connection" for HTTP transport is referring to the duration of the session across polling requests which ends when the Agent restarts or explicitly reconnects. Updated the wording to clarify this.

specification.md
liustanley: Added a new section "Certificate Hierarchy and Extended Key Usage" to clarify this

specification.md
liustanley: TLS server certificates carry `id-kp-serverAuth`, which will fail the EKU check because the Agent requires `id-kp-codeSigning` during chain validation. This means that TLS certificates cannot be repurposed as signing certificates. Edited this paragraph to clarify this.

supplementary-guidelines.md
liustanley: Added an ASCII sequence diagram that illustrates this flow and also represents the previous diagram

specification.md
liustanley: Removed the "empty" wording here

specification.md
liustanley: > We can look into adding sequence_num to ServerToAgent to prevent the replay attacks.

Should we scope this to a separate PR or address it in this one?

tigrannajaryan: Let's do in a separate PR. sequence_num is also interesting for other reasons, and I want to have that discussed separately. There is a bit of nuance there (e.g. how are the server instances expected to generate sequential numbers in a horizontally scaled deployment) that warrants its own discussion.

Once we have sequence_num we can update attestation section to make use of it.

proto/opamp.proto
```diff
@@ -244,6 +244,19 @@ message ServerToAgent {
     // A custom message sent from the Server to an Agent.
     // Status: [Development]
     CustomMessage custom_message = 11;
+
+    // Field numbers 12 and 13 were briefly assigned to inline payload
+    // trust verification metadata (trust_chain_response and signature)
+    // in an earlier draft of the Message Attestation spec. The design
```
tigrannajaryan: Look like there are some comments which leak the history of this PR originating on a fork. Can you please clean it so that whatever it says makes sense in the current context (this is the upstream, and the history of the fork doesn't matter here).

specification.md
liustanley: I agree that a "readonly" mode would be an interesting fallback mechanism, but I'm inclined to leave this mechanism as out of scope for the initial message attestation spec. We are already defining message attestation as an opt-in capability, so users who opt-in will likely be interfacing with servers that support attestation and will expect full read-write capabilities.

If a "readonly" mode is desired in the future, it may make sense to propose this as a separate spec from message attestation, since it could apply to users who are opt-out as well.

specification.md
liustanley: > The support of TOFU flow for obtaining attestation trust anchor can be a configuration option in the Agent.

Supporting a TOFU flow via a configuration option sounds reasonable with me.

My main concern with using a TOFU flow for message attestation is that persistent storage is required in order to persist the acquired trust anchor across restarts. If the Agent is deployed in a container environment for example, the Server connection can be ephemeral due to container restarts, which would result in repeated TOFU flows while reconnecting.

Even with these drawbacks in mind, I think it's still reasonable to support TOFU as an optional, simpler startup flow, but we should note these drawbacks for customers to know.

liustanley:
Hi @tigrannajaryan, I've finished addressing the comments on this PR and opened the following prototype and example PRs on the opamp-go repository:
- https://github.com/open-telemetry/opamp-go/pull/599
- https://github.com/open-telemetry/opamp-go/pull/600

Please let me know if anything else is required for review, thanks in advance

specification.md
```diff
@@ -3561,6 +3592,671 @@ recommend the following:
...
+### Algorithm
+
+The signing algorithm is determined by the leaf certificate's public key (subjectPublicKeyInfo), paired with the hash function specified for that key type in the list below. The OpAMP protocol does not negotiate algorithms.
+
+Implementations SHOULD support, at minimum, the following algorithms:
+
+* ECDSA P-256 with SHA-256
+* ECDSA P-384 with SHA-384
+* RSA-2048 or larger with PKCS#1 v1.5 and SHA-256
+* Ed25519
```
andykellr, line 4192: From Claude: 
> ECDSA signature encoding is unspecified — a guaranteed interop failure. “ECDSA P-256 with SHA-256” doesn’t say ASN.1 DER (Go’s ecdsa.SignASN1, OpenSSL default) or raw fixed-width r‖s (JWS/COSE/WebCrypto default). Two conforming implementations will not interoperate. Pin one; DER fits an X.509-centric design. Related: PKCS#1 v1.5 is listed and PSS is not — say whether PSS is forbidden or merely not required.

The `opamp-go` implementation comment says DER-encoded but we should be explicit.

specification.md
```diff
@@ -3561,6 +3592,671 @@ recommend the following:
...
+  not confidentiality.
+* **Authentication of `AgentToServer` messages.** Verifying the
+  authenticity of Agent-originated messages is tracked separately in
+  [opamp-spec issue #20](https://github.com/open-telemetry/opamp-spec/issues/20).
+* **Trust anchor distribution by the Server.** The payload trust anchor
+  is operator-managed and MUST NOT be modified by any field of any
+  `ServerToAgent` message.
+* **Algorithm negotiation.** The certificate's `signatureAlgorithm` is
+  authoritative. Future algorithm support is added by certificate
+  issuers and X.509 stacks, not by changes to OpAMP.
```
andykellr: From Claude:
> Algorithm and Out of Scope contradict each other. Algorithm says the signing algorithm comes from the leaf’s subjectPublicKeyInfo. Out of Scope says “the certificate’s signatureAlgorithm is authoritative.” Different fields — signatureAlgorithm describes the issuer’s signature over the cert and says nothing about the subject key. The SPKI rule is right; fix the bullet.

specification.md
```diff
@@ -3561,6 +3592,671 @@ recommend the following:
...
+The certificate chain (intermediates plus leaf) MUST chain to the
+pre-configured payload trust anchor. The trust anchor itself is
+supplied out-of-band and MUST NOT be included in the
+`certificate_chain` field of `trust_chain_response`.
+
+Revocation checking is RECOMMENDED. Implementations SHOULD use the
+revocation-checking facilities of their X.509 library (CRL distribution
+points, OCSP) during chain validation. Operators MAY rely on
+short-lived certificates as an alternative to active revocation.
+
```
andykellr: From Claude:
> The revocation SHOULD is dead text in the reference language. The spec recommends using “the revocation-checking facilities of their X.509 library (CRL distribution points, OCSP)” and lists Go’s crypto/x509 as a covering stack. x509.Certificate.Verify performs no revocation checking whatsoever — there is no facility to enable. Since opamp-go is the reference implementation, either drop this in favor of the short-lived-certificate guidance already present, or state that revocation is out of band.

specification.md
```diff
@@ -3561,6 +3592,671 @@ recommend the following:
...
+When the payload trust anchor approaches expiry or must be replaced,
+rotation is handled by the same out-of-band mechanism used to provision
+it initially (for example, updating the certificate file in the Agent's
+configuration and restarting the Agent). Operators SHOULD plan for this
+operational burden, such as using a long-lived root CA or automating
+certificate distribution through their existing configuration-management
+tooling. The same out-of-band rotation applies to TOFU-enrolled anchors:
+to rotate, delete the persisted anchor file and restart the Agent.
+
+### Trust On First Use (TOFU)
```
andykellr, line 3705: From Claude:
> TOFU looks half-landed. Beyond the substantive objection (for the common case where the agent is installed by a script fetched from the server being trusted, TOFU buys nearly nothing for real complexity), it is not internally consistent:
> - It contradicts the Trust Model’s own sentence and the still-unedited Out of Scope bullet saying the anchor “MUST NOT be modified by any field of any ServerToAgent message.”
> - tofu_trust_anchor (field 3) is in proto/opamp.proto but absent from the TrustChainResponse block in specification.md.
> - AcceptsPayloadTrustAnchorTOFU = 0x00020000 is in the proto but absent from the AgentCapabilities enum in specification.md.
> - Neither appears in the CHANGELOG, which still describes only RequiresPayloadTrustVerification.
> - The TOFU section is missing from the table of contents.

andykellr:
I added some comments some my review discussion with Claude. Rather than attempt to restate these concerns, I just referenced its feedback directly.

specification.md
liustanley: Added a paragraph after the algorithm list to specify the signature encoding per key type:
- ECDSA -> MUST be ASN.1 DER `SEQUENCE {r, s}`; raw `r || s` MUST NOT be used.
- RSA -> MUST use RSASSA-PKCS1-v1_5 with SHA-256; RSASSA-PSS MUST NOT be used
- Ed25519 -> RFC 8032 PureEdDSA, encoding fixed by that spec

specification.md
liustanley: Edited this paragraph to state that revocation is out of band and that short-lived signing certificates are a primary mitigation for a compromised signing key.

andykellr (approved):

specification.md
liustanley: Opened an issue to address the sequence_num change: https://github.com/open-telemetry/opamp-spec/issues/398

PR #334, opened 2026-05-12
Add ComponentHealth.attributes

### LINKED ISSUE #326: Add attributes to ComponentHealth messages
We should extend the definition of [ComponentHealth messages](https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#componenthealth-message) to include attributes.

For example:

```proto
// The health of the Agent and sub-components
// Status: [Beta]
message ComponentHealth {
    // Set to true if the component is up and healthy.
    bool healthy = 1;

    // Timestamp since the component is up, i.e. when the component was started.
    // Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January 1970.
    // If the component is not running MUST be set to 0.
    fixed64 start_time_unix_nano = 2;

    // Human-readable error message if the component is in erroneous state. SHOULD be set
    // when healthy==false.
    string last_error = 3;

    // Component status represented as a string. The status values are defined by agent-specific
    // semantics and not at the protocol level.
    string status = 4;

    // The time when the component status was observed. Value is UNIX Epoch time in
    // nanoseconds since 00:00:00 UTC on 1 January 1970.
    fixed64 status_time_unix_nano = 5;

    // A map to store more granular, sub-component health. It can nest as deeply as needed to
    // describe the underlying system.
    map<string, ComponentHealth> component_health_map = 6;

    // Additional context or metadata for the observed component's status
    repeated KeyValue attributes = 7;
}
```

Attributes are a part of the componentstatus.Event structs that the [healthcheck extensions use](https://github.com/open-telemetry/opentelemetry-collector/blob/8ca992c832690e561f7fe70563ac48ec8c37bb0e/component/componentstatus/status.go#L87-L96). Extending the spec to allow these attributes to be passed will make it easier for clients that use the healthcheck extension to send the entire component status through OpAMP

michel-laterman:
Add `attributes` in order to be consistent with what the healthcheck extension produces. See [extension source](https://github.com/open-telemetry/opentelemetry-collector/blob/8ca992c832690e561f7fe70563ac48ec8c37bb0e/component/componentstatus/status.go#L87-L96).

- Closes #326

### DIFF
--- a/proto/opamp.proto
+++ b/proto/opamp.proto
@@ -812,6 +812,7 @@ message ComponentHealth {
     map<string, ComponentHealth> component_health_map = 6;
 
     // Additional context or metadata for the observed component's status.
+    // Attributes are component specific and outside the concerns of the OpAMP protocol.
     repeated KeyValue attributes = 7;
 }
 
--- a/proto/opamp.proto
+++ b/proto/opamp.proto
@@ -810,6 +810,9 @@ message ComponentHealth {
     // A map to store more granular, sub-component health. It can nest as deeply as needed to
     // describe the underlying system.
     map<string, ComponentHealth> component_health_map = 6;
+
+    // Additional context or metadata for the observed component's status.
+    repeated KeyValue attributes = 7;
 }
 
 message EffectiveConfig {
--- a/proto/opamp.proto
+++ b/proto/opamp.proto
@@ -813,6 +813,7 @@ message ComponentHealth {
 
     // Additional context or metadata for the observed component's status.
     // Attributes are component specific and outside the concerns of the OpAMP protocol.
+    // Status: [Development]
     repeated KeyValue attributes = 7;
 }
 
--- a/specification.md
+++ b/specification.md
@@ -72,6 +72,7 @@ Status: [Beta]
       - [ComponentHealth.status](#componenthealthstatus)
       - [ComponentHealth.status_time_unix_nano](#componenthealthstatus_time_unix_nano)
       - [ComponentHealth.component_health_map](#componenthealthcomponent_health_map)
+      - [ComponentHealth.attributes](#componenthealthattributes)
     + [EffectiveConfig Message](#effectiveconfig-message)
       - [EffectiveConfig.config_map](#effectiveconfigconfig_map)
     + [RemoteConfigStatus Message](#remoteconfigstatus-message)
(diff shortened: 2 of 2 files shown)


tigrannajaryan (changes requested):
What are "attributes" supposed to carry? We need a more elaborate motivation before accepting this change.

Looking at the extension event, it says "attributes provides additional context or metadata for the event" which is very vague to me.

michel-laterman:
For example the attributes of a hostmetrics scraper can include things to indicate specifics on which scrapers are failing, as seen in the [extension issue](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/43606) is:
```json
    "pipeline:metrics/1": {
      "healthy": true,
      "status": "StatusRecoverableError",
      "status_time": "2025-10-16T17:56:11.820542393+02:00",
      "attributes": {
        "error_msg": "not enough permissions to read cpu data",
        "scrapers": ["cpu", "memory", "network"]
      },
```
[extension implementation pr](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/43744).

It could further be used to indicate other metadata such as port numbers, address, etc.

proto/opamp/v1/opamp.proto
```diff
@@ -810,6 +810,9 @@ message ComponentHealth {
     // A map to store more granular, sub-component health. It can nest as deeply as needed to
     // describe the underlying system.
     map<string, ComponentHealth> component_health_map = 6;
+
+    // Additional context or metadata for the observed component's status.
```
tigrannajaryan, line 815: Can we add that the attributes are agent-type specific to make it clear OpAMP has no opinion about the content?

tigrannajaryan (approved):
@michel-laterman sorry for late review. I think it makes sense for us to mirror the structure of the status that the Collector has.

tigrannajaryan:
I will keep this open for a while in case @andykellr or @evan-bradley have comments.

proto/opamp/v1/opamp.proto
```diff
@@ -810,6 +810,10 @@ message ComponentHealth {
...
+    // Attributes are component specific and outside the concerns of the OpAMP protocol.
+    repeated KeyValue attributes = 7;
```
tigrannajaryan, line 818: Please mark with `// Status: [Development]` (see other similar fields in this proto).

tigrannajaryan:
Please resolve the conflict.

andykellr (approved):

PR #350, opened 2026-06-03
Add role to AgentConfigFile

### LINKED ISSUE #184: Add a field for non-config files to AgentRemoteConfig
### Summary

When configuring an agent, it may be helpful to provide files to the agent that are specifically designated as files that are related to the agent's configuration, but are not configuration files themselves.

### Context

This change could be applicable to any agent, but is currently motivated by the way the Collector works. The Collector allows specifying multiple configuration files passed through its command line arguments, and will merge the resulting files. For example:

```
otelcol --config=file0.yaml --config=file1.yaml --config=file2.yaml
```

The Collector also allows specifying pulling in configuration from other sources within config files defined in command line arguments. For example, say `file0.yaml` above contains the following line:

```
rules: ${file:ruleset.yaml}
```

The contents of `ruleset.yaml` would be substituted into the contents of `file0.yaml` before the full configuration is parsed by the Collector.

If we want to provide multiple files to the Collector through the command line, but also provide files that are read inside of the Collector's configuration files, we would need to specify the difference between files that are top-level configuration and those that are supplementary to configuration. This could include config fragments as above, but also other files such as certificates, binary files, etc. that may be read by the Collector on startup.

#### Add a new field in `ServerToAgent`

We could add another `AgentRemoteConfig`-typed field on `ServerToAgent` named `remote_supplementary_config` or similar that contains files intended to aid those in `remote_config`.

#### Add a new field in `AgentRemoteConfig`

We could add another list of files in `AgentRemoteConfig` that contains supplementary configuration files.

#### Add a new field in `AgentConfigFile` that specifies the role of a file

We could add another string field, e.g. `config_type` on `AgentConfigFile` that specifies how the server intends for the file to be used by the agent, and rely on semantic conventions for what these types mean. This would be the most flexible approach, but would also require the most processing by the agent.

#### Using `PackagesAvailable` to provide supplementary files

We could realistically offer supplementary files as packages, but this would be out-of-band for configuration and would likely be a complicated flow.

#### Utilize the MIME type of config files

We could alter the MIME type of config files, such as adding a parameter, to include OpAMP-specific information that indicates to the agent how a file should be used. This would also be a flexible option, but would need careful consideration to ensure that the OpAMP annotations don't cause unintended conflicts.

dmitryax:
Adds optional `role` to `AgentConfigFile` so agent-specific implementations can distinguish top-level configuration from supplementary content.

Fixes #184

### DIFF
--- a/proto/opamp/v1/opamp.proto
+++ b/proto/opamp/v1/opamp.proto
@@ -1045,13 +1045,18 @@ message AgentConfigMap {
 }
 
 message AgentConfigFile {
-    // Config file or section body. The content, format and encoding depends on the Agent
-    // type. The content_type field may optionally describe the MIME type of the body.
+    // Config file, section, or supplementary content body. The content, format and
+    // encoding depends on the Agent type. The content_type field may optionally
+    // describe the MIME type of the body.
     bytes body = 1;
 
     // Optional MIME Content-Type that describes what's in the body field, for
     // example "text/yaml".
     string content_type = 2;
+
+    // Optional role of the content in the body field. The values and their semantics
+    // are Agent type-specific.
+    string role = 3;
 }
 
 /////////////////////////////////////////////////////////////////////////////////////
--- a/specification.md
+++ b/specification.md
@@ -2626,6 +2626,7 @@ following structure:
 message AgentConfigFile {
   bytes body = 1;
   string content_type = 2;
+  string role = 3;
 }
 ```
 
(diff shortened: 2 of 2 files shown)


tigrannajaryan:
This looks good to me.

tigrannajaryan:
@dmitryax do you want to make this ready for review?

dmitryax:
@tigrannajaryan sorry, I missed this PR. Marked as ready for review

tigrannajaryan (approved):

tigrannajaryan:
I will keep it open for a bit in case @andykellr or @evan-bradley want to review.

RassK:
The description is a bit vague, I did not get the exact point.
The first part is very generic:
> The values and their semantics are Agent type-specific.

The name itself suggests some fixed purpose.

Why not then just `metadata` instead of a `role`, so it's much more compatible with wider needs?

tigrannajaryan:
@dmitryax are we good to go or you planning to address [this comment](https://github.com/open-telemetry/opamp-spec/pull/350#issuecomment-4865624869)?

Issue #338, opened 2026-05-16
Restructure proto folders to align with versioned packages


juandemanjon:
With the introduction of versioned packages (`opamp.proto.v1`), we should establish a clear directory structure that aligns with future versioning needs.

**Current Structure:**
- `proto/` folder with package `opamp.proto.v1`

**Proposed Structure:**
- `proto/opamp/v1/` → package `opamp.v1` (removing proto in the package name, breaking change)
- `proto/opamp/v2/` → package `opamp.v2` (future, consistent package naming)

**Testing & Validation**

This restructuring has already been tested in separate branches:
- **opamp-spec**: Tested with the new proto folder structure
- **opamp-go/internal/opamp-spec**: Go code was successfully generated in the `protobufs` folder with minimal changes to the Makefile

```makefile
BASELINE_PROTO_FILES := $(wildcard internal/opamp-spec/proto/opamp/v1/*.proto)
```

In conclusion, the proposed structure is compatible with existing code generation workflows.

PR #352, opened 2026-06-03
Restructure proto folders: move from proto to proto/opamp/v1

juandemanjon:
# Proto Restructuring
## Overview
This PR introduces changes to the proposed #338 proto folder structure.

### 1. Proto Folder Restructuring
- **Move proto files** from `proto/` to `proto/opamp/v1/`
- **Update package name** from `opamp.proto.v1` to `opamp.v1`

**Outcome**: aligns with standard proto package naming conventions

### 2. Configuration Updates
- **Updated** `Makefile` to reflect new proto file paths, and generate Go, Python, and C# code
- **Updated** `specification.md` to reference the new proto structure

## Breaking Changes
- Python code is sensitive to the new package structure and needs to be updated
- C# namespace has changed to `Opamp.V1`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,18 +51,21 @@ markdownlint:
 
 GENDIR := gen
 # Find all .proto files.
-PROTO_FILES := $(wildcard proto/*.proto)
+PROTO_FILES := $(wildcard proto/opamp/v1/*.proto)
 
 PROTO_GEN_GO_DIR ?= $(GENDIR)/go
+PROTO_GEN_PYTHON_DIR ?= $(GENDIR)/python
+PROTO_GEN_CSHARP_DIR ?= $(GENDIR)/csharp
 
-OTEL_DOCKER_PROTOBUF ?= otel/build-protobuf:0.14.0
+# https://github.com/open-telemetry/build-tools/releases 
+OTEL_DOCKER_PROTOBUF ?= otel/build-protobuf:0.25.0
 
 # Docker pull image.
 .PHONY: docker-pull
 docker-pull:
 	docker pull $(OTEL_DOCKER_PROTOBUF)
 
-gen-proto: gen-go
+gen-proto: gen-go gen-python gen-csharp
 .PHONY: gen-proto
 
 # Generate Protobuf Go files.
@@ -72,6 +75,28 @@ gen-go:
 	mkdir -p ./$(PROTO_GEN_GO_DIR)
 
 	# Verify generation of Go protos
-	$(foreach file,$(PROTO_FILES),$(call exec-command,docker run --rm -v${PWD}:${PWD} \
+	$(foreach file,$(PROTO_FILES),$(call exec-command,docker run --rm -u $(shell id -u):$(shell id -g) -v${PWD}:${PWD} \
         -w${PWD} $(OTEL_DOCKER_PROTOBUF) --proto_path=${PWD}/proto/ \
         --go_out=./$(PROTO_GEN_GO_DIR) -I${PWD}/proto/ ${PWD}/$(file)))
+
+# Generate Protobuf Python files.
+.PHONY: gen-python
+gen-python:
+	rm -rf ./$(PROTO_GEN_PYTHON_DIR)
+	mkdir -p ./$(PROTO_GEN_PYTHON_DIR)
+
+	# Verify generation of Python protos
+	$(foreach file,$(PROTO_FILES),$(call exec-command,docker run --rm -u $(shell id -u):$(shell id -g) -v${PWD}:${PWD} \
+        -w${PWD} $(OTEL_DOCKER_PROTOBUF) --proto_path=${PWD}/proto/ \
+        --python_out=./$(PROTO_GEN_PYTHON_DIR) -I${PWD}/proto/ ${PWD}/$(file)))
+
+# Generate Protobuf C# files.
+.PHONY: gen-csharp
+gen-csharp:
+	rm -rf ./$(PROTO_GEN_CSHARP_DIR)
+	mkdir -p ./$(PROTO_GEN_CSHARP_DIR)
+
+	# Verify generation of C# protos
+	$(foreach file,$(PROTO_FILES),$(call exec-command,docker run --rm -u $(shell id -u):$(shell id -g) -v${PWD}:${PWD} \
+        -w${PWD} $(OTEL_DOCKER_PROTOBUF) --proto_path=${PWD}/proto/ \
+        --csharp_out=./$(PROTO_GEN_CSHARP_DIR) -I${PWD}/proto/ ${PWD}/$(file)))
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -72,6 +75,28 @@ gen-go:
...
+# Generate Protobuf Python files.
+.PHONY: gen-python
+gen-python:
+	rm -rf ./$(PROTO_GEN_PYTHON_DIR)
+	mkdir -p ./$(PROTO_GEN_PYTHON_DIR)
+
+	# Verify generation of Python protos
+	$(foreach file,$(PROTO_FILES),$(call exec-command,docker run --rm -u $(shell id -u):$(shell id -g) -v${PWD}:${PWD} \
+        -w${PWD} $(OTEL_DOCKER_PROTOBUF) --proto_path=${PWD}/proto/ \
+        --python_out=./$(PROTO_GEN_PYTHON_DIR) -I${PWD}/proto/ ${PWD}/$(file)))
```
tigrannajaryan: We don't have anything Python-specific in the proto, so I suggest we remove this section. We can keep gen-csharp since we have csharp-specific package names in proto.

juandemanjon: Done!

Makefile
```diff
@@ -51,18 +51,21 @@ markdownlint:
 
 GENDIR := gen
 # Find all .proto files.
-PROTO_FILES := $(wildcard proto/*.proto)
+PROTO_FILES := $(wildcard proto/opamp/v1/*.proto)
```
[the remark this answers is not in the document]
juandemanjon: This requires versioned output directories (`gen/v1/`, `gen/v2/`, etc.), which is an architectural concern outside this PR's scope. The current setup outputs everything to a single `gen` location

tigrannajaryan:
@RassK @pellared any concerns with this change? I assume not an issue for your language binding generation?

proto/opamp/v1/anyvalue.proto
```diff
@@ -20,10 +20,10 @@
 
 syntax = "proto3";
 
-package opamp.proto.v1;
+package opamp.v1;
 
 option go_package = "github.com/open-telemetry/opamp-go/protobufs";
-option csharp_namespace = "OpAmp.Proto.V1";
+option csharp_namespace = "OpAmp.V1";
```
tigrannajaryan: @RassK please check this is correct.

proto/opamp/v1/opamp.proto
```diff
@@ -16,12 +16,12 @@
...
 
-package opamp.proto.v1;
+package opamp.v1;
 
-import "anyvalue.proto";
+import "opamp/v1/anyvalue.proto";
 
 option go_package = "github.com/open-telemetry/opamp-go/protobufs";
-option csharp_namespace = "OpAmp.Proto.V1";
+option csharp_namespace = "OpAmp.V1";
```
tigrannajaryan: @RassK and here.

RassK:
> @RassK @pellared any concerns with this change? I assume not an issue for your language binding generation?

Just generates breaking changes and backwards incompatibility - but in the same time, it's anyway experimental feature (so breaking changes should be acceptable).

The more questionable part is if the package name change is necessary. Since then it's not consistent with OpenTelemetry any more and follows different convention. Every single other package is using `proto` in it's name.

juandemanjon:
> > @RassK @pellared any concerns with this change? I assume not an issue for your language binding generation?
> 
> Just generates breaking changes and backwards incompatibility - but in the same time, it's anyway experimental feature (so breaking changes should be acceptable).
> 
> The more questionable part is if the package name change is necessary. Since then it's not consistent with OpenTelemetry any more and follows different convention. Every single other package is using `proto` in it's name.

Thanks for raising this. I want to advocate for keeping OpAMP structurally independent from OTel.

My understanding is that the `.proto.` namespace distinction was intentionally introduced to differentiate the namespaces across various language packages and specifically to prevent collisions with SDKs, APIs, and other components.

Architecturally, OpAMP has a completely different footprint and purpose than OpenTelemetry's core components. In my opinion, maintaining strict independence and avoiding tight coupling or structural dependencies with OTel is vital for OpAMP's long-term flexibility and adoption across diverse agent ecosystems.

tigrannajaryan:
> The more questionable part is if the package name change is necessary.

Go bindings only use the last part of the package name, so it doesn't change anything there.

If this is not the case for other languages we should not make this change.

I welcome a clean up, but not if it causes a breaking change. If it impacts consumers and creates work for them we should not do this, it is not worth it.

juandemanjon:
> The more questionable part is if the package name change is necessary. 

The package name change is not a problem because it was [changed recently](https://github.com/open-telemetry/opamp-spec/commit/a642d8b8e88e828ef6c70054d361c6e9dbd80fd0) from `package opamp.proto;` to `package opamp.proto.v1;` and nobody raised any concern.

To be clear, the possible issue is the new folder structure, which, in some languages, could result in the protobuf code generator output location being different (the code itself does not change), and the downstream components will need to change their imports.

RassK:
.NET was using the lately commited namespace ahead of time (since it was missing anyway), that's why there were no issues.

Now it's a different story... but also as I said it's not a big deal. So I'm not going to block but for users / vendors (if there are any besides OTel) it will be just annoying a bit.

proto/opamp/v1/anyvalue.proto
```diff
@@ -20,10 +20,10 @@
 
 syntax = "proto3";
 
-package opamp.proto.v1;
```
tigrannajaryan, line 23: We should follow the approach OTLP takes, which uses `proto` in the package name: https://github.com/open-telemetry/opentelemetry-proto/blob/41ad255e3efd5450096fc167ef6d87edb4bcb4f3/opentelemetry/proto/trace/v1/trace.proto#L17

tigrannajaryan (dismissed):
Blocking this since it diverges from what we do in OTLP.

proto/opamp/v1/opamp.proto
juandemanjon: No problem, I can restore .proto. in the package name

tigrannajaryan:
@RassK PTAL. From what I see this should have no impact on generated code since it just changes local directory structure without touching package names.

RassK (approved):

tigrannajaryan (approved):

tigrannajaryan:
I will keep this open for a while in case @andykellr has any comments.

juandemanjon:
@tigrannajaryan Thank you for merging it.

There is a follow-up PR in opamp-go to support the new proto location

https://github.com/open-telemetry/opamp-go/pull/573

chalin:
Hi @tigrannajaryan et all: please give @open-telemetry/docs-maintainers a heads up when there are org changes like these so that we can make adjustments in the otel.io side of things. Thanks!

tigrannajaryan:
@chalin I will. Sorry, I didn't realize it will impacts the docs.

chalin:
> @chalin I will. Sorry, I didn't realize it will impacts the docs.

Thanks. It turned out not to be as bad as I thought 😅. So no worries, this probably won't happen often.

PR #364, opened 2026-06-30
docs: update OpAMP implementations list

cijothomas:
Update the implementations list with OpenTelemetry-owned OpAMP clients and Collector components.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,8 +66,10 @@ For more information about the emeritus role, see the [community repository](htt
 
 ### Libraries
 
-- [OpAMP Go](https://github.com/open-telemetry/opamp-go)
+- [OpAMP Go Client and Server](https://github.com/open-telemetry/opamp-go)
 - [OpAMP .NET Client](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/main/src/OpenTelemetry.OpAmp.Client)
+- [OpAMP Java Client](https://github.com/open-telemetry/opentelemetry-java-contrib/tree/main/opamp-client)
+- [OpAMP Python Client](https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/opamp/opentelemetry-opamp-client)
 
 ### Agent Management Platforms
 


README.md
```diff
@@ -66,15 +66,12 @@ For more information about the emeritus role, see the [community repository](htt
...
-- [Bindplane](https://bindplane.com)
+- [OpAMP Java Client](https://github.com/open-telemetry/opentelemetry-java-contrib/tree/main/opamp-client)
+- [OpAMP Python Client](https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/opamp/opentelemetry-opamp-client)
 
 ### Agents
 
-- [Bindplane Distro for OpenTelemetry Collector](https://github.com/observIQ/bindplane-otel-collector)
-- [OpAMP Extension for OpenTelemetry Collector](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/opampextension)
-- [OpAMP Supervisor for OpenTelemetry Collector](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/cmd/opampsupervisor)
+- [OpenTelemetry Collector Contrib OpAMP Extension](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/opampextension)
```
cijothomas: @tigrannajaryan @andykellr I removed the Bindplane link given it is not OpenTelemetry owned product. Happy to revert and keep it back, if that was intentional.

README.md
```diff
@@ -66,15 +66,12 @@ For more information about the emeritus role, see the [community repository](htt
 
 ### Libraries
 
-- [OpAMP Go](https://github.com/open-telemetry/opamp-go)
+- [OpAMP Go client and server](https://github.com/open-telemetry/opamp-go)
 - [OpAMP .NET Client](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/main/src/OpenTelemetry.OpAmp.Client)
-
-### Agent Management Platforms
-
-- [Bindplane](https://bindplane.com)
```
tigrannajaryan, line 74: I would like to keep a reference to Bindplane since it is the only known production grade OpAMP server implementation as far as I know. We want to have a good reference point for people who want to explore OpAMP for production use.

If we get to a point where OpAMP is better supported we can decide whether we want to remove non-OpenTelemetry assets from this list.

@andykellr if there is a better link we can use please suggest an update.

cijothomas: Without a more specific link, there is nothing a user can go check for themselves.

Also see https://github.com/open-telemetry/opentelemetry.io/pull/10306

tigrannajaryan: Sure, we should correct the fact on whether it is OSS or no.

Regardless, I would like to have a link to the only known fully functional agent management system using OpAMP. I believe it is valuable to have it here.

cijothomas: left Bindplane as-is for this PR.

RassK (approved):

tigrannajaryan (approved):

README.md
tpaschalis: Grafana Fleet Management has also gone GA with its OpAMP capabilities. I've opened a PR to add it to the list of supported platforms if that's OK.

PR #381, opened 2026-08-02
Fix SDK service namespace identifying attribute

cijothomas:
## Summary

- Change the SDK identifying attribute from `service.namespace.name` to `service.namespace`.

## Rationale

`service.namespace.name` is not an OpenTelemetry semantic convention attribute. The service resource conventions define `service.namespace`, and the service-instance identity is the `service.namespace`, `service.name`, and `service.instance.id` tuple. This also aligns the SDK guidance with the Collector guidance in this document.

The incorrect name was introduced in #312 while translating the service identity tuple into the SDK-specific OpAMP guidance.

## Validation

- `markdownlint`

### DIFF
--- a/opentelemetry-guidelines.md
+++ b/opentelemetry-guidelines.md
@@ -48,7 +48,7 @@ Language instrumentation agents MUST copy the following OpenTelemetry SDK
 
 - `service.name`
 - `service.instance.id`
-- `service.namespace.name`, if present
+- `service.namespace`, if present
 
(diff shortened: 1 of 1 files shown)


tigrannajaryan (approved):

tigrannajaryan:
Looks like a typo to me, added in https://github.com/open-telemetry/opamp-spec/pull/312
FYI @breedx-splk

PR #385, opened 2026-08-03
Clarify configuration map semantics: rename "file" to "object", allow empty key unconditionally

assafad1:
While reviewing open-telemetry/opentelemetry-collector-contrib#49692 (adding `reports_raw_config` to the OTel Collector), I noticed the spec's "Configuration Files" section has a couple of things that don't match how the protocol is actually used.

The empty-key restriction doesn't make sense. The spec says `""` is only allowed when there's a single entry in the map, but there's no good reason for that - it's just a convention. @tigrannajaryan confirmed it was unintentional.

"File" is also the wrong abstraction. The entries in `AgentConfigMap` aren't files, they're named blobs. The contrib PR is a good example: it uses `""` for the expanded config and a separate key for an alternative representation. Neither is a file.

What changed:
- Renamed `AgentConfigFile` to `AgentConfigObject` (suggested in SIG call)
- Renamed the section from "Configuration Files" to "Configuration Objects" and updated prose throughout
- Keys are now explicitly agent-implementation-defined; any key including `""` is valid
- Added a note that multiple keys can represent alternative views of the same configuration
- Dropped "raw" from the body field description
- Fixed a pre-existing typo: "AgentConfigSet" -> "AgentConfigMap"

Follows up on open-telemetry/opentelemetry-collector-contrib#49692 with @douglascamata, @jade-guiton-dd, and @tigrannajaryan.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v0.20.0
+
+* Rename `AgentConfigFile` to `AgentConfigObject` by @assafad1 in https://github.com/open-telemetry/opamp-spec/pull/385
+
 ## v0.19.0
 
 * Add transport message size limits by @pellared in https://github.com/open-telemetry/opamp-spec/pull/346
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## v0.20.0
 
-* Rename `AgentConfigFile` to `AgentConfigObject` by @assafad1 in https://github.com/open-telemetry/opamp-spec/pull/385
+* **Breaking change**: Rename `AgentConfigFile` to `AgentConfigObject` by @assafad1 in https://github.com/open-telemetry/opamp-spec/pull/385
 
 ## v0.19.0
 
--- a/proto/opamp/v1/opamp.proto
+++ b/proto/opamp/v1/opamp.proto
@@ -1036,11 +1036,10 @@ message AgentRemoteConfig {
 }
 
 message AgentConfigMap {
-    // Map of configs. Keys are config file names or config section names.
-    // The configuration is assumed to be a collection of one or more named config files
-    // or sections.
-    // For agents that use a single config file or section the map SHOULD contain a single
-    // entry and the key may be an empty string.
+    // Map of configuration objects. Keys are agent-implementation-defined names
+    // for each object. An Agent type MAY assign special meaning to specific keys,
+    // including the empty string (""). For agents that use a single configuration
+    // object the map SHOULD contain a single entry and the key MAY be an empty string.
     map<string, AgentConfigFile> config_map = 1;
 }
 
--- a/proto/opamp/v1/opamp.proto
+++ b/proto/opamp/v1/opamp.proto
@@ -1040,10 +1040,10 @@ message AgentConfigMap {
     // for each object. An Agent type MAY assign special meaning to specific keys,
     // including the empty string (""). For agents that use a single configuration
     // object the map SHOULD contain a single entry and the key MAY be an empty string.
-    map<string, AgentConfigFile> config_map = 1;
+    map<string, AgentConfigObject> config_map = 1;
 }
 
-message AgentConfigFile {
+message AgentConfigObject {
     // Config file, section, or supplementary content body. The content, format and
     // encoding depends on the Agent type. The content_type field may optionally
     // describe the MIME type of the body.
--- a/proto/opamp/v1/opamp.proto
+++ b/proto/opamp/v1/opamp.proto
@@ -1038,8 +1038,9 @@ message AgentRemoteConfig {
 message AgentConfigMap {
     // Map of configuration objects. Keys are agent-implementation-defined names
     // for each object. An Agent type MAY assign special meaning to specific keys,
-    // including the empty string (""). For agents that use a single configuration
-    // object the map SHOULD contain a single entry and the key MAY be an empty string.
+    // including the empty string (""). The empty string is a valid key and MAY be
+    // used regardless of how many entries are in the map. For agents that use a
+    // single configuration object the map SHOULD contain a single entry.
     map<string, AgentConfigObject> config_map = 1;
 }
 
--- a/specification.md
+++ b/specification.md
@@ -2612,7 +2612,7 @@ message:
 
 ```protobuf
 message AgentConfigMap {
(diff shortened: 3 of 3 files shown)


specification.md
tigrannajaryan: I think we should also rename this to `AgentConfigObject`. This will not break the protocol on the wire. It will be a breaking change for generated Protobuf bindings, but I think that's acceptable, since we are still in Beta.

@andykellr @evan-bradley what do you think?

This is also related to discussion in https://github.com/open-telemetry/opamp-spec/issues/184

cc @dmitryax @pavankrish123

tigrannajaryan (commented):
I support this change. We should discuss this in the OpAMP SIG meeting tomorrow.

specification.md
tigrannajaryan: Discussed in SIG call today. Let's rename the message.

specification.md
```diff
@@ -2613,13 +2616,13 @@ message AgentConfigMap {
...
+The config_map field of the AgentConfigMap message is a map where keys are
+agent-implementation-defined names for each configuration object.
 
-For Agents that use a single config file the config_map field SHOULD contain a
-single entry and the key MAY be an empty string.
+For Agents that use a single configuration object the config_map field SHOULD
+contain a single entry.
 
-The AgentConfigFile message represents one configuration file and has the
+The AgentConfigFile message represents one configuration object and has the
```
tigrannajaryan:
```suggestion
The AgentConfigObject message represents one configuration object and has the
```

proto/opamp/v1/opamp.proto
```diff
@@ -1036,11 +1036,10 @@ message AgentRemoteConfig {
...
-    // Map of configs. Keys are config file names or config section names.
-    // The configuration is assumed to be a collection of one or more named config files
-    // or sections.
-    // For agents that use a single config file or section the map SHOULD contain a single
-    // entry and the key may be an empty string.
+    // Map of configuration objects. Keys are agent-implementation-defined names
+    // for each object. An Agent type MAY assign special meaning to specific keys,
+    // including the empty string (""). For agents that use a single configuration
+    // object the map SHOULD contain a single entry and the key MAY be an empty string.
     map<string, AgentConfigFile> config_map = 1;
```
tigrannajaryan:
```suggestion
    map<string, AgentConfigObject> config_map = 1;
```

hitkall: looks good to me

assafad1: Done

specification.md
assafad1: Done

specification.md
assafad1: Done

assafad1:
Hey @tigrannajaryan @hitkall, I've addressed your comments. Please let me know if anything else is needed from my side :)

proto/opamp/v1/opamp.proto
```diff
@@ -1036,15 +1036,14 @@ message AgentRemoteConfig {
...
-    map<string, AgentConfigFile> config_map = 1;
+    // Map of configuration objects. Keys are agent-implementation-defined names
+    // for each object. An Agent type MAY assign special meaning to specific keys,
+    // including the empty string (""). For agents that use a single configuration
+    // object the map SHOULD contain a single entry and the key MAY be an empty string.
+    map<string, AgentConfigObject> config_map = 1;
 }
 
-message AgentConfigFile {
+message AgentConfigObject {
```
tigrannajaryan, line 1047: Please add an entry in CHANGELOG.md (for future version 0.20.0) saying AgentConfigFile is renamed to AgentConfigObject, with references to this PR number.

tigrannajaryan (commented):
LGTM, just one comment.

proto/opamp/v1/opamp.proto
assafad1: Done

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## v0.20.0
+
+* Rename `AgentConfigFile` to `AgentConfigObject` by @assafad1 in https://github.com/open-telemetry/opamp-spec/pull/385
```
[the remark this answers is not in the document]
assafad1: fixed

specification.md
```diff
@@ -2590,48 +2590,51 @@ Config │    Config │  ServerToAgent{AgentRemoteConfig} │   │and     │
...
   string content_type = 2;
   string role = 3;
 }
 ```
 
-The body field contains the raw bytes of the configuration file. The content,
-format and encoding of the raw bytes is Agent type-specific and is outside the
+The body field contains the bytes of the configuration object. The content,
+format and encoding of the body bytes is Agent type-specific and is outside the
 concerns of OpAMP protocol.
```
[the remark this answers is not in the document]
assafad1: fixed

proto/opamp/v1/opamp.proto
```diff
@@ -1036,15 +1036,14 @@ message AgentRemoteConfig {
...
-    // Map of configs. Keys are config file names or config section names.
-    // The configuration is assumed to be a collection of one or more named config files
-    // or sections.
-    // For agents that use a single config file or section the map SHOULD contain a single
-    // entry and the key may be an empty string.
-    map<string, AgentConfigFile> config_map = 1;
+    // Map of configuration objects. Keys are agent-implementation-defined names
+    // for each object. An Agent type MAY assign special meaning to specific keys,
+    // including the empty string (""). For agents that use a single configuration
+    // object the map SHOULD contain a single entry and the key MAY be an empty string.
```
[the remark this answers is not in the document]
assafad1: fixed

evan-bradley (approved):

andykellr (approved):

juandemanjon:
This is a breaking change in the protos.

We need to decide when breaking changes won't be allowed anymore.

There are best practices to avoid breaking changes, such as marking the existing field as deprecated and adding a new one

```protobuf

map<string, AgentConfigFile> config_map = 1 [deprecated = true];

map<string, AgentConfigObject> config_object_map = 2;
```

douglascamata:
@juandemanjon: the spec is considered beta, so breaking changes that are judged necessary can be made. 

The good thing is that this doesn’t break the protocol on the wire, but only generated code because of the renaming. This should be easily detectable because, depending on the language, code just won’t compile. 

We discussed this in the previous SIG call and the general agreement there was that we are fine with this change.

PR #387, opened 2026-08-05
Update operations sequence description

### LINKED ISSUE #366: Spec clarification required - Message order & transport locks
### What

Many operations in OpAMP require somekind of message order (described in the [operation](https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#operation) section). 

HTTP transport:
> When the Agent wants to send a message to the Server and the Agent has previously sent a request to the Server that is not yet responded, the Client MUST wait until the response is received before a new request can be made. Note that the new request in this case can be made immediately after the previous response is received, the Client does not need to wait for the polling period between requests.

Many things happen async in the background, that can interrupt the sequence:
- Heartbeats (interval)
- Effective config (on agent load, after changes)
- Remote config apply (async for applying and applied state)
- Custom async capabilities
- etc

So my question is:
- Should client enforce message queue to maintain the sequence?
- Should server be able to respond in async manner? when sequence is interrupted 
- When server responds with requests that requires async answer (like report full state, apply remote config, etc), should other events (Heartbeats, etc listed previously) wait in the queue?
- Anything more important to add? Http vs WebSockets etc?

My feeling in implementing the client is that some parts expect more of a sync manner but actual life runs in async.

RassK:
### Why

Fixes #366

### What

Describes a corner case when messages accumulate while the pipe is blocked. Specifies how both (client & server) sides should act not to cause data loss or dropped packages.

### DIFF
--- a/specification.md
+++ b/specification.md
@@ -21,6 +21,7 @@ Status: [Beta]
 
 - [Introduction](#introduction)
 - [Communication Model](#communication-model)
+  * [Message Sequences](#message-sequences)
   * [WebSocket Transport](#websocket-transport)
     + [WebSocket Message Format](#websocket-message-format)
     + [WebSocket Message Size Limits](#websocket-message-size-limits)
@@ -363,6 +364,19 @@ In the rest of the OpAMP specification the term _Agent_ is used to refer to the 
 which implements the client portion of the OpAMP, regardless of whether or not a
 Supervisor is part of that entity.
 
+### Message Sequences
+
+The message sequences described in this specification are logical sequences.
+Multiple logical sequences MAY be in progress at the same time and their
+messages MAY be interleaved. Unless a particular operation explicitly says
+otherwise, a message that is sent as part of one logical sequence is not required
+to be the next message sent after the message that triggered it.
+
+The sequence is normally started by an initiating message triggered by some
+external event. For example after the connection is established the Client sends
+an AgentToServer message. In this case the "connection established" is the
+triggering event and the AgentToServer is the initiating message.
+
 ### WebSocket Transport
 
 One of the supported transports for OpAMP protocol is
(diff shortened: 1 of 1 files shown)


specification.md
```diff
@@ -465,6 +465,12 @@ OpAMP over WebSocket is an asynchronous, full-duplex message exchange protocol.
 sequence of messages exchanged by the OpAMP Client and the Server is defined for each
 particular capability in the corresponding section of this specification.
 
+The message sequences described in this specification are logical sequences.
```
tigrannajaryan: This is currently in "WebSocket Message Exchange", but it is applicable to both transports. Can we move this under the parent "Communication Model" section, perhaps make it a new subsection of "Communication Model", called something like "Message Sequences".

The description itself is very good.

specification.md
```diff
@@ -541,6 +547,12 @@ when using the WebSocket transport. The only difference is in the timing:
   this case can be made immediately after the previous response is received, the Client
   does not need to wait for the polling period between requests.
 
+While an HTTP request is in progress the Client MAY accumulate changes that need
+to be sent to the Server. After the HTTP response is received the Client MAY send
+an accumulated AgentToServer message immediately, even if processing the
+ServerToAgent message received in the response later produces additional data to
+report.
```
tigrannajaryan, line 557: 👍

tigrannajaryan (commented):
Mostly LGTM, just one comment about the placement of the text.

specification.md
RassK: I have added "Message Sequences" section. Grabbed a bit more than I added but it feels like a lot from the `WebSocket Message Exchange` can be moved there? - eg until heartbeats are mentioned.

https://github.com/open-telemetry/opamp-spec/pull/387/commits/9f53dad03cdce23d823b6f66f8a9ac5e441d5177

RassK: I detected one exception here. When client initializes the connection and server responds with initial information, then client must wait AND process the message before any accumulation can even take place.

The reason is simple, client gathers capabilities and only then the client can say (send) something.

And now there is another exception coming from:

```
enum ServerCapabilities {
    // The Server can accept status reports. This bit MUST be set, since all Server
    // MUST be able to accept status reports.
    AcceptsStatus                  = 0x00000001;
}
```

So clients have every right to assume, that status is accepted and accumulated status report could be sent before processing the server reply.

RassK:
@tigrannajaryan could you take a look again ⬆️

specification.md
RassK: It came up that this accumulator logic might corrupt custom messages when client has no idea if the message was sent or accumulated. Wondering if this should be addressed in spec or clients should provide feedback in some form.

tigrannajaryan (approved):

specification.md
tigrannajaryan: That's a good point. I don't know what the right solution is. I see a few possibilities:
- Return an error if a custom message is pending and the client tries to set another custom message.
- Have a queue of custom messages to be sent.
- Change `custom_message` field to be an array in proto, thus allowing to accumulate multiple within a single AgentToServer mesage.
- Something else?

tigrannajaryan: @andykellr what do you think?

RassK: I don't have preferences here. Each layer could solve it but the only issue is if spec does not do anything, then this bug easily propagates to different languages.

Kielek:
@tigrannajaryan, @andykellr, can it be merged?

Kielek (approved):

PR #396, opened 2026-09-08
Update GnuPG detached signature links

Kielek:
gnupg changed the layout of the pages. It includes anchor within the pages. 160--->161 is intentional


Kielek:
Done in #393

